Issue #49, opened 2014-07-31
why " disable keep-alives -- not needed with Go 1.3 "


andybrier:
in proxy/proxy.go:
                         #disable keep-alives -- not needed with Go 1.3
            responseWriter.Header().Set("Connection", "close")
anyone can explain why not needed with go 1.3? . the code above results int 2 connection header in http response.

PR #75, opened 2015-02-14
Allow keepalives to the front-end proxy

fraenkel:
- Improve https traffic by 4x
- Improve http traffic by 20%
- Closes the gap between http and https traffic


thsandfuchs:
relates to #49

PR #81, opened 2015-04-20
Add job_name to component.announce messages

voelzmo:
If the property 'name' in the config is empty, the message stays the same.


jpalermo:
We've decided not to merge these pull requests in. They do provide a piece of data the is currently missing, but the number of changes that will be required to the jobs and their templates seems to great considering this functionality is already planned for inclusion in loggregator/metron/firehose components.

PR #93, opened 2015-08-10
Log the contents of X-Forwarded-Proto

CAFxX:
We need X-Forwarded-Proto to appear in the access logs (cfr. http://docs.pivotal.io/pivotalcf/opsguide/ssl-term.html#other)

### DIFF
--- a/access_log/access_log_record.go
+++ b/access_log/access_log_record.go
@@ -55,6 +55,7 @@ func (r *AccessLogRecord) makeRecord() *bytes.Buffer {
 	fmt.Fprintf(b, `"%s" `, r.FormatRequestHeader("User-Agent"))
 	fmt.Fprintf(b, `%s `, r.Request.RemoteAddr)
 	fmt.Fprintf(b, `x_forwarded_for:"%s" `, r.FormatRequestHeader("X-Forwarded-For"))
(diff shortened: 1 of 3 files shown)


simonjohansson:
@CAFxX this should be of interest to you :)

CAFxX:
JSON :scream:

shinji62:
@simonjohansson the X-Forwarded-Proto is pairing with the  X-Forwarded-For so not really strange to add here instead of extra headers..

simonjohansson:
@shinji62 ah yes, fair enough. :)

CAFxX:
So... can this go in? :)

shalako:
Thanks for the contribution. Seems like a reasonable enhancement. Prioritizing review.

CAFxX:
As a follow-up I was thinking to check with pprof if it might be a win to get rid of some Fprintfs, but that will have to wait for #92 to be merged. 

Looking forward to have this one merged. Thanks!

fraenkel:
There are ways to improve the building of the log string but the biggest hurt right now is that we do it twice, once for the access log and the other for loggregator.

CAFxX:
@fraenkel sure will look into that as a followup to this and #92

CAFxX:
Looks like this was merged in 55cf98781198bea091dbe3022ab5078cf3379da8. Closing.

CAFxX:
@fraenkel BTW, as anticipated above the followup is in #96.

PR #96, opened 2015-09-11
Optimize access_log_record.go

CAFxX:
Coalesce log line printing in a single Sprintf and memoize the result so that we don't build each log line twice
Three small functional changes:
- Missing values are replaced by `-` instead of the verbose `Missing<name-of-field>` placeholder
- The request duration does not print pointless 0s after the decimal separator
- Extra headers values are properly escaped using strconv.Quote

With the changes to the benchmark included in the PR (github.com/cloudfoundry/gorouter/access_log/file_and_loggregator_access_logger_test.go) we get, without the patch:
```
  Ran 500 samples:
  writeTime:
    Fastest Time: 0.009s
    Slowest Time: 0.046s
    Average Time: 0.011s ± 0.005s
```

and, with the patch:
```
  Ran 500 samples:
  writeTime:
    Fastest Time: 0.004s
    Slowest Time: 0.036s
    Average Time: 0.005s ± 0.003s
```

### DIFF
--- a/access_log/access_log_record.go
+++ b/access_log/access_log_record.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
@@ -21,6 +22,7 @@ type AccessLogRecord struct {
 	BodyBytesSent        int
 	RequestBytesReceived int
 	ExtraHeadersToLog    []string
+	record               string
 }
 
 func (r *AccessLogRecord) FormatStartedAt() string {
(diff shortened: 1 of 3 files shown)


shalako:
Hello @CAFxX, thank you for your contribution. Was this discussed on cf-dev first? In the future, please ping me with a proposal before doing the work. Thanks.

- Please provide example log output for before and after
- I'm skeptical that removing `Missing<name of field>` is a positive change. This format is self-documenting and more immediately meaningful. What is driving this change? Some performance improvement by the removal of a few characters? If we are going to go forward with this change, we'll need discoverable documentation for the log format, that describes the order and meaning of fields in the log line. This should be added here: https://github.com/cloudfoundry/gorouter#logs

CAFxX:
@shalako it wasn't discussed on cf-dev, but briefly discussed with @fraenkel in #92.

The primary driver was simply optimizing the code without changing any existing behavior. But during the refactoring I noticed some issues:

- Most fields, when missing, are replaced by a `-`. Just three of them are replaced by `Missing<NameOfField>`. For consistency I replaced them with `-`. This change is opinionated and can be reverted, if you think it is appropriate.
- The request duration field was formatted with 9 digits after the decimal separator in all cases. So 60 would be formatted as `60.000000000`, 60.1 would be `60.100000000` and 60.000000001 would be `60.000000001`. I changed it to print decimal digits only if they are meaningful (i.e. `60`, `60.1` and `60.000000001` for the cases above). Again, opinionated change that can be reverted if needed.

How about I rescope this PR for the optimization part (i.e. no format changes), and leave the opinionated changes for another day?

shalako:
@CAFxX Thank you for the clarification. 

No need to rescope yet. I'll pull the story into the backlog to have the team review.

access_log/access_log_record.go
```diff
@@ -39,49 +41,56 @@ func (r *AccessLogRecord) ResponseTime() float64 {
...
-	fmt.Fprintf(b, `x_forwarded_for:"%s" `, r.FormatRequestHeader("X-Forwarded-For"))
-	fmt.Fprintf(b, `x_forwarded_proto:"%s" `, r.FormatRequestHeader("X-Forwarded-Proto"))
-	fmt.Fprintf(b, `vcap_request_id:%s `, r.FormatRequestHeader("X-Vcap-Request-Id"))
-
-	if r.ResponseTime() < 0 {
-		fmt.Fprintf(b, "response_time:MissingFinishedAt ")
-	} else {
-		fmt.Fprintf(b, `response_time:%.9f `, r.ResponseTime())
+	if r.ResponseTime() >= 0 {
+		responseTime = strconv.FormatFloat(r.ResponseTime(), 'f', -1, 32)
```
leochu: Response time includes precisions up to nanoseconds which can be up to 9 digits after the period.
We would like to avoid losing precision in the access log for the response time field.

A sample case when the above code would lose precision in the response time field: 
http://play.golang.org/p/6hiqyDR37h

CAFxX:
@leochu good catch. Fixed.

msmykowski:
@CAFxX We were curious why you decided to write the access log record twice in the benchmark test?

https://github.com/rakutentech/gorouter/blob/4d67340cfc283ceec37cae61c8099aff42b67533/access_log/file_and_loggregator_access_logger_test.go#L85-L88

crhino:
This PR was pulled in on c068f24e4089f6807c389a072cf0d8767db6ac66, but wasn't merged so no automatic closing of this PR happened.

Closing now.

PR #98, opened 2015-09-23
Emit router metrics through dropsonde

roxtar:
This change is required for router metrics to be available through the loggregator firehose.

- Moved ProxyReporter from proxy.go to the metrics package
- The RouteRegistry takes in an additional RouteReporter parameter
  which records route stats


leochu:
Would it make sense for the CompositeReport to contain a list of ProxyReporter instead of limiting to two?

roxtar:
@leochu We had thought about creating a slice of reporters, but we couldn't think of any case where we would have more than two. At the end we agreed that if we were to add one more reporter, to get the total reporters to be three, we would change it to a slice. Thoughts?

shalako:
@roxtar how should I validate the change? I can see some results using: 
```
$ DOPPLER_ADDR=wss://doppler.a1.cf-app.com:4443 CF_ACCESS_TOKEN="$(cf oauth-token | grep bearer)" bin/firehose_sample | grep router
```
If I get any results does that demonstrate the change?

roxtar:
@shalako Yes, looking at the firehose is one way. I created a [datadog dashboard](https://app.datadoghq.com/dash/71640?live=true&page=0&is_auto=false&from_ts=1443751814620&to_ts=1443755414620&tile_size=m) comparing the router metrics coming from the firehose and collector.

The latency metric emitted to the collector is always hardcoded to 5e-7. The one emitted on the firehose is the actual latency.

The last two graphs don't line up exactly due to the differences in how metrics are published in the two mechanism. Loggregator uses the push model for publishing metrics, where the component is responsible for emitting metrics at the frequency of their choice. With the collector the metrics are pulled at an interval so when metrics remain constant then those values are also published.

leochu:
@roxtar We're okay with not passing in a slice for now.  Thanks!

shalako:
@roxtar I don't see uptime in the output for 

```
$ DOPPLER_ADDR=wss://doppler.a1.cf-app.com:4443 CF_ACCESS_TOKEN="$(cf oauth-token | grep bearer)" bin/firehose_sample | grep router
```

How can I validate this through the firehouse?
Is there a way to inspect metron directly, to see if this metric is making it that far?

roxtar:
@shalako It looks like that the uptime metric was not present when the PR was submitted but added recently. Looking at [the commit](https://github.com/cloudfoundry/gorouter/commit/6fae6d0a0e856b70d30234904c671a0e3be64cdc) it doesn't look like that metric is being sent to loggregator.

shalako:
These commits were merged Sept 30. Closing.

Thank you, @roxtar and @ebabani !

Issue #103, opened 2015-11-24
Data race running gorouter


fraenkel:
Running tests caught the following:

==================
WARNING: DATA RACE
Write by goroutine 15:
  github.com/cloudfoundry-incubator/routing-api.(*client).SetToken()
      /Users/fraenkel/t/src/github.com/cloudfoundry-incubator/routing-api/client.go:48 +0x33
  github.com/cloudfoundry/gorouter/route_fetcher.(*RouteFetcher).FetchRoutes()
      /Users/fraenkel/t/src/github.com/cloudfoundry/gorouter/route_fetcher/route_fetcher.go:114 +0x125
  github.com/cloudfoundry/gorouter/route_fetcher.(*RouteFetcher).StartFetchCycle.func1()
      /Users/fraenkel/t/src/github.com/cloudfoundry/gorouter/route_fetcher/route_fetcher.go:47 +0x8f

Previous write by goroutine 16:
  github.com/cloudfoundry-incubator/routing-api.(*client).SetToken()
      /Users/fraenkel/t/src/github.com/cloudfoundry-incubator/routing-api/client.go:48 +0x33
  github.com/cloudfoundry/gorouter/route_fetcher.(*RouteFetcher).subscribeToEvents()
      /Users/fraenkel/t/src/github.com/cloudfoundry/gorouter/route_fetcher/route_fetcher.go:72 +0x18c
  github.com/cloudfoundry/gorouter/route_fetcher.(*RouteFetcher).StartEventCycle.func1()
      /Users/fraenkel/t/src/github.com/cloudfoundry/gorouter/route_fetcher/route_fetcher.go:60 +0x2e

shalako:
Hi, @fraenkel could you provide config used to run the tests so we can reproduce? Errand or manual? Manifest or config file snippets would be helpful.

Thank you

fraenkel:
just ran scripts/test with go 1.5.1
See https://travis-ci.org/cloudfoundry/gorouter/jobs/92951383

shalako:
Pulling this in and calling it a chore as there doesn't appear to be any user-facing changes.

shalako:
Closing as this was fixed.

PR #104, opened 2015-11-24
Update dropsonde

fraenkel:
- Dropsonde emits one event HttpStartStop for each request.
- A request metric will always be emitted rather than potentially dropped if
  it exceeded 1 minute.

- Move flag parsing into main to allow current Go versions to work


fraenkel:
Go 1.3 should be removed from Travis.
Go 1.5 and higher fails because of the race as mentioned #103

shalako:
Could you please provide acceptance criteria for these changes?

- Dropsonde emits one event HttpStartStop for each request.
If this is a change, what is the current behavior?

- A request metric will always be emitted rather than potentially dropped if
it exceeded 1 minute.
If the request exceeds 1 minute, a metric isn't currently emitted?

- Move flag parsing into main to allow current Go versions to work
What is considered "current"? 1.4 and 1.5? How is this validated from an operators perspective (create release / deploy)?

fraenkel:
Dropsonde used to send an HttpStart and HttpStop event. Metron would cache the HttpStart and match the HttpStop if the HttpStart was still available and send an HttpStartStop. The new Dropsonde code just has a single event emitted, HttpStartStop.
The HttpStart was only kept for 1 minute in Metron and no HttpStartStop was sent if there was no match on the HttpStop.

See any travis build on Go tip which has the same issue with Go 1.5.x. Ginkgo/go test fails.
see https://travis-ci.org/cloudfoundry/gorouter/jobs/92984949
func init() is meant for packages to initialize themselves prior to main() running. It is poor practice to have one in main since you are in control.

shalako:
Is this correct?
Given an HTTP request that takes longer than a minute
When the request closes I should see an HttpStartStop message in the firehose
(Currently I will see no message in the firehose)

The change to main() looks like an implementation detail; I'll leave that to the team to review.

fraenkel:
Correct, you will always see an event for every request that completes rather than those that only take < 1 minute.

shalako:
Thank you. Pulling the story in for review.

crhino:
@fraenkel Does the updated dropsonde require go1.5.1? If so we will need to update the golang version in cf-release as well.

fraenkel:
No, just what I am using locally.

PR #107, opened 2015-12-02
Fixes #106 - Allow HTTP request target in absolute-form

### LINKED ISSUE #106: Route lookup not working for HTTP requests in absolute-form
We faced an issue with certain HTTP clients talking to CF, which sent requests with a request target in absolute-form. While this form is meant to be used when talking to proxies, if I understand the spec (https://tools.ietf.org/html/rfc7230#section-5.3.2) properly, the servers also must support it.

Here is an example:
```
GET http://my.test.io/my-path HTTP/1.1
Host: my.test.io
```
results in "404 Not Found: Requested route ('my.test.io') does not exist." in a setup in which the following would work.
```
GET /my-path HTTP/1.1
Host: my.test.io
```

The issue seems to be in the lookup() func in proxy.go, where host and requestURI get concatenated. In the case described above this would result in a lookup for the first two parts of something like my.test.iohttp://my.test.io/my-path

This may be a rather rare case, still I'd like to provide a patch in case you also see it as an issue. I'll submit a PR, which you can review.

krumts:
If the router recieves a request like
  GET http://my.test.io/my-path HTTP/1.1
  Host: my.test.io

an error message "Requested route ('my.test.io') does not exist." is returned, even though the route my.test.io exists.

This patch provides a test and fix for the issue.

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -124,7 +124,14 @@ func (p *proxy) getStickySession(request *http.Request) string {
 }
 
 func (p *proxy) lookup(request *http.Request) *route.Pool {
-	uri := route.Uri(hostWithoutPort(request) + request.RequestURI)
+	var requestPath string
+	if request.URL.IsAbs() {
+		requestPath = request.URL.Path
+	} else {
+		requestPath = request.RequestURI
+	}
+
+	uri := route.Uri(hostWithoutPort(request) + requestPath)
 	return p.registry.Lookup(uri)
 }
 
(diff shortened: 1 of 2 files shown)


shalako:
Closing for now; see #107

markstgodard:
Reviewed and tested locally. Looks good. We will be merging your PR. 

-Mark and @shashwathi 
(CF Routing team)

crhino:
a9044d1f87988250ec3a69b26a83fdc9a8d49735

shalako:
@markstgodard @crhino How come this PR was closed and not merged?

markstgodard:
@shalako This PR has been merged in. The PR was not automatically mergable since there were changes in this area since the PR was created, so we rebased and merged the author's changes in.

Issue #105, opened 2015-11-27
Collecting infrastructure access log records


stoyanr:
Hi,

We would like to be able to collect the access log records of the infrastructure components into our ELK stack. Currently this is not possible. As can be seen in https://github.com/cloudfoundry/gorouter/blob/68aff9a593eb565cef23906cb8c678189de6b9d1/access_log/create_running_access_logger.go and https://github.com/cloudfoundry/gorouter/blob/68aff9a593eb565cef23906cb8c678189de6b9d1/access_log/file_and_loggregator_access_logger.go:

* the access log records are written to a file, if a file is specified in config/gorouter.yml ("access_log").
* the access log records are also sent via dropsonde, but only if the app_id is non-empty

Therefore, infrastructure access log records (with empty app_id) are not sent anywhere besides the file.

One idea how to achieve this would be to add a new (optional) entry in config/gorouter.yml for these log records, e.g. `infra_access_log: <%= p("router.infra_access_log") %>`. If this entry is present, log records with empty app_id are sent in addition to a different file (if specified) or to stdout (if a file is not specified). From stdout, they are sent to syslog via the log capture mechanism in common/utils.sh, and can be collected from there.

Does this make sense? Would you accept a pull request with such a change?

stoyanr:
I submitted a new pull request for this issue, [PR #108](https://github.com/cloudfoundry/gorouter/pull/108). Instead of the proposal above, we thought it would be better to introduce general support for sending access log records with empty app_id to the syslog. For details, see the description in the pull request itself.

shalako:
Closing in favor of #108

PR #108, opened 2015-12-02
Added support for sending access log records to syslog

stoyanr:
As described in [Issue 105](https://github.com/cloudfoundry/gorouter/issues/105), we at SAP would like to be able to collect the gorouter access log records into our ELK stack. Currently, access log records with empty app_id are not sent anywhere besides the file configured via the `access_log` configuration setting.

The change introduces support for sending access log records to the syslog. By default it is not active, so the current behavior is preserved. It can be activated and configured via the following additional configuration settings:

    access_log_syslog:
      enabled: true # default is false
      syslog: vcap.gorouter.access_log # syslog prefix, default is empty
      level: warn # syslog logging level, default is debug
      app_id_filter: all # empty, non_empty, or all, default is all

Note that for the sake of consistency, the same values for `level` are supported as in the `logging` section. These are converted to their corresponding syslog levels when configuring the syslog logger.

The `app_id_filter` setting controls whether to send the access log records to the syslog based on their app_id. The possible values are:
* `empty` - only records with empty app_id are sent to the syslog
* `non_empty` - only records with non-empty app_id are sent to the syslog
* `all` - all records are sent to the syslog


shalako:
Hello,

I'm supportive of adding optional syslog streaming of gorouter logs, but 1) I'm surprised this isn't already supported, 2) we need to explore what the current convention is for other components in cf (https://www.pivotaltracker.com/story/show/109685796), and 3) your solution seems overly specific.

I would approach this feature as: "When property foo is enabled, gorouter logs are streamed to the specified syslog consumer, at the log level specified by existing property router.logging_level." 

- We already have a property for log level `router.logging_level`. This should apply for local logs as well as syslog streaming.
- Syslog streaming should apply for more than just `access_control.log`. I'm thinking about `gorouter.log` and `drain.log`.

Thank you for your contribution. Stand by while we do some investigation.

stoyanr:
Hi Shannon,

Thanks for your reply. In fact, syslog streaming of gorouter log is already supported, but only for the logs of the gorouter itself (`gorouter.log`), not for access log records. It is implemented using steno in a way consistent with other CF components. Our solution is specific to the access logs since these logs are already treated in a rather special way - they are not logged via steno (and consequently not streamed to the syslog), they are written to a separate file (`access.log`), and in addition access log records with a non-empty app_id are sent via dropsonde to the loggregator. Therefore, we believe that the proposed change fits well into the existing design. We could have simply sent the access logs to the steno logger, but this would have led to having them in both log files, something we thought is not desirable, and we couldn't find a way to configure steno in two different ways for two different groups of logs in the same component.

The property you are referring to (which is in fact `logging.level` in the gorouter configuration) controls the level at which gorouter log records are written by steno at all. The level with which these records are streamed to the syslog is hardcoded as "debug" in https://github.com/cloudfoundry/gosteno/blob/master/syslog_sink.go. The `access_log_syslog.level` in our implementation controls exactly this second level for the access log records, and therefore it is not the same as `logging.level`, and cannot be replaced by it.

We would be fine with any solution that meets the following requirements:
* Access log records are streamed to the syslog
* It is possible to configure which log records are streamed based on the app_id, i.e. we would like to be able to stream only infra logs (app_id is empty), only application logs (app_id is non empty), or both (the `access_log_syslog.app_id_filter` setting above) 

Ideally, we would also like to be able to distinguish between gorouter logs and access logs based on their prefix (the `access_log_syslog.syslog` setting above), and also specify the syslog logging level (the `access_log_syslog.level` setting above).

We would be happy to correct the pull request in accordance with the results of your investigation.

CAFxX:
+1 from us at Rakuten

crhino:
Note: The travis CI failure is do to golang 1.3 not being supported by gnatsd. Since we currently run the GoRouter on 1.4.3, we have removed it from the travis.yml

shalako:
My preference would be that access logs are forwarded to syslog in the same way that gorouter logs are, and that the same log level applies to all logs, whether they are written locally or streamed to syslog. If `logging.level` doesn't control the log level of syslog forwarding, it should. An additional property seems superfluous. 

I'm concerned about adding the suggested filtering in gorouter for access logs. It seems reasonable for gorouter to send all access logs; filtering should be done by an external syslog aggregator. 

We're still discussing these topics as a team. We'll get back to you when we have more information.

stoyanr:
Hi Shannon, 

As far as I understand, the following approach would fit better for you: 
* The same `logging` configuration that controls the syslog forwarding of gorouter logs applies to the access logs. This means that these logs are always sent to the syslog, with the same prefix and level as the gorouter logs. 
* Access logs are written locally to a different file (`access_log`). As with gorouter logs, if a file is not specified, they are sent to the console.
* It is not possible to forward to the syslog only application or only infrastructure logs, all access logs are always forwarded.

We could rework the pull request, just please let us know if the proposed approach really fits. 

> If logging.level doesn't control the log level of syslog forwarding, it should. An additional property seems superfluous.

These are 2 different things:
* `logging.level` controls which log entries are actually logged by steno (to all sinks), e.g. if it is configured as "info" and the code logs something as "debug", this entry is not logged. This doesn't really apply to access logs, as they have no level.
* The level with which any steno logs are actually forwarded to the syslog is hardcoded (in steno) as "debug". This is a property of the steno sink and not easily changed by individual logging calls.

As this is an orthogonal concern compared to the initial intention of the pull request, I would propose not to try to address it with this pull request. So for the time being, access logs would be forwarded to the syslog as "debug", just as gorouter logs (and likely logs of all other components that use steno) currently are.

shalako:
@stoyanr Yes, this is what I was thinking. Thank you for the additional explanation re `logging.level`. Before submitting a change to your PR, I want to discuss your proposal with our team. Please stand by.

shalako:
@stoyanr 

Your latest proposal sounds good. We'd welcome your PR with the changes we've discussed.

As the location of gorouter.log and access.log are both in the job template, it does not seem like this PR requires any new manifest properties. Is this correct?

For each PR I like to understand the acceptance criteria. This is how I imagine testing the change; please correct me if I misunderstand:
1. Once cf-release branch release-candidate includes the bump to gorouter that includes the proposed changes, `git checkout release-candidate`, `bosh create release`, `bosh upload release`
2. Update the manifest property `syslog_daemon_config` with a syslog consumer (e.g. papertrailapp.com)
3. `bosh deploy`
4. Verify that the same content available in `gorouter.log` and `access.log` can be found in the syslog consumer

stoyanr:
@shalako Yes, since we will remove all additional configuration settings and use the ones already available in the steno configuration, the PR will not require any new manifest properties. The acceptance criteria that you described is also how we are going to test it.

stoyanr:
@shalako We adapted the pull request as discussed, and tested it as you described. Please check to see if it fits your expectations.

There are few small differences in what we implemented to what we proposed above:
* If an access log file is not specified, the access logs are not sent to stdout, for consistency with the previous behavior and to avoid clutter when gorouter logs are also sent there
* If the `logging.syslog` setting is empty, the access logs are not sent to the syslog, for consistency with gorouter logs.

Therefore, with the default configuration the access logs are not sent anywhere, as before the change.

ishustava:
Hi @stoyanr,

Sorry we have taken so long to get back to you. In the beginning of the new year, we decided to switch the GoRouter away from the `gosteno` logging framework and towards the [`cf-lager` framework](https://github.com/cloudfoundry-incubator/cf-lager) used by the vast majority of other Cloud Foundry go projects. This will make logging more consistent and allow us to potentially use libraries such as [cicerone](https://github.com/cloudfoundry-incubator/cicerone) that other teams have developed.

With this change, we also considered how to satisfy your requirements for sending access logs through syslog. What we ended up settling on was this:
* Keep both `gorouter.log` and `access.log` for the separate local log files
* If access log file is not configured, than no access logs file will be written **and** no syslog messages sent
* If gorouter log file is not configured, the local file will not be written to, **but** log messages will still be sent to stdout (and the gorouter_ctl.log file).
* By sending access logs through stdout when configured, the [GoRouter ctl script](https://github.com/cloudfoundry/cf-release/blob/develop/jobs/gorouter/templates/gorouter_ctl.erb#L7) sources the `common/utils.sh` file, which will send stderr and stdout to syslog.

This differs slightly from your PR, notably around merging both access logs and gorouter logs into the stdout stream, and always sending access logs to syslog when the access_log file is configured. We do not believe that clutter inside of the `gorouter_ctl.log` is a big issue since the separate files are still there locally, and when searching in ELK filtering can be applied. For the `logging.syslog` configuration, moving to `cf-lager` obsoleted that configuration value and gorouter logs are always sent to syslog.

We would like to know if these changes have satisfied your requirements. If not, please let us know and we can look into it.

@ishustava && @crhino, CF Routing Team

stoyanr:
Hi @ishustava && @crhino,

Thanks for replying. I think the changes are ok from our perspective (if I understand them correctly, see below). Is everything already committed as described? Could we already check if it works for us?

I have a couple of questions:
* If an access log file is configured, then the logs are also sent to syslog. And if not, then the logs are again sent to syslog via the ctl script, so they would reach syslog in both cases, correct?
* You mention that the `logging.syslog` configuration is obsolete. It was used before to specify the syslog "prefix" (e.g. `vcap.gorouter`) used to identify gorouter logs. Without it, is this prefix now hardcoded?

I assume this particular PR is not relevant anymore and should be closed, correct?

atulkc:
Hi @stoyanr,

The `steno` to `cf-lager` changes are now committed. We are working on getting the `logging.syslog` prefix back. With regards to your questions:

* If access log is not configured, the logs are not sent to syslog.
* As we have mentioned earlier, we missed the prefix configuration and will bring it back, so the previous functionality will remain the same.

You are correct, this PR is no longer relevant. We will close it when all the changes are committed.

@atulkc and @ishustava, CF Routing Team

atulkc:
Hi @stoyanr 

Please use following commit https://github.com/cloudfoundry/cf-release/commit/9537ce42565c1781eb059154816c37e75f142881 to test out syslog changes.

I am closing this PR as all the changes related to syslog are now done.

Thanks,
@atulkc, CF Routing team

shalako:
Hello, @stoyanr 
As reported in #119 we've seen a significant performance impact by streaming access logs. We're going to disable streaming access logs by default, then add an manifest property to re-enable this feature; @youngm has contributed a solution in #117.

Issue #117, opened 2016-03-04
Questions about access logs in stdout


youngm:
I recently noticed this commit: https://github.com/cloudfoundry/gorouter/commit/f079b7a64cf35bc3a5b7f5f06f7557b07a951c2a

We may be weird but in my organization we make monitoring data including component logs public to our users.  We recently noticed the access logs of other applications started appearing in our public component logs.  I think some of our users and uncomfortable with other users seeing their application access logs.

That said I also really like this commit because it allows me to see CC and other non application access logs in my central monitoring logs so I'm a little torn on this commit.

At the very least would you consider a config option to turn off sending access logs to stdout?  At the most it would be great if I could continue to get access logs for routes that don't have an app Guid.  But I can see how that may not be gorouters job.

Thoughts?

shalako:
Hi Mike,

I am surprised that you make the logs of CF system components to end users. I haven't heard of that use case before. 

As you're probably aware, we send access logs to stdout because we use lager to stream all logs out over syslog. If we were to add a property, it would be something like `syslog_includes_access_log:true` (open to better suggestion for naming). I'd want the property to control whether access logs were streamed to syslog; the use of stdout and lager is an implementation detail. Make sense?

Could you send us PR for this? 

Thank you

youngm:
@shalako Understood. I'll put together a PR.  Thanks Shannon!

youngm:
@shalako I think I'd be fine for our needs with turning access logs completely off on the gorouter.  Thoughs on simply submitting a PR to allow the `access_log` config to be customized in the release?  Then I can just give it an empty value and disable access logging that way?  Or perhaps you'd prefer an explicit flag to turn off access logging?

shalako:
FYI, the team is digging into our use of lager base on performance reports in #119. There may be some changes in that area of the code as a result. I believe we're still testing at this point.

Please help me understand the difference between a manifest property that can be used to turn off streaming of access log to syslog (you could implement this by turning off sending access log to stdout; I don't care), and your proposal to allow config of `access_log`? Do you mean that access logs would also not be logged locally? ...when on - log locally and stream to syslog via stdout; when off - do neither...

Thank you

youngm:
[...]

Issue #119, opened 2016-03-08
replacing gosteno with lager causing huge performance degradation


milkotodorov:
Hi @crhino ,

I've found that since commit "replacing gosteno with cf_lager" f97f07c13c79e60a0048975e9639a99e8310ad0a there is a huge performance degradation regarding the requests that can be processed by the gorouter (about 10x times and more in some cases). In this commit there is a replacement of the existing gosteno logging library with the lager one, which seems to cause this performance slowdown.

The following diagrams are all made on AWS - gorouter is running on a c3.2xlarge VM (8x CPU/15GB RAM). I use an application which can handle directly on a single instance about 4000-4500 req/sec and it is scaled to 4 instances. And nGrinder to generate as a loadgenerator. Each request is 1024 bytes and the app replies with 1024 bytes back. 

With gorouter v229 commit 8f63d53079e572cbfc1cfba9c07c58499a280224 sending the requests directly to the gorouter the result looks like this:
![v229_8f63d53](https://cloud.githubusercontent.com/assets/14890886/13608313/8442c5a2-e554-11e5-968c-5fb2abcdb0a8.png)

The same look like if I try to send a fake requests to the router (to an non-existing app), so it can simply reply with 404. 

However with gorouter v229 commit f97f07c13c79e60a0048975e9639a99e8310ad0a the same scenario looks like this:

![v229_f97f07c](https://cloud.githubusercontent.com/assets/14890886/13608746/73afa1f4-e556-11e5-8792-4fc22c22266c.png)

By the way in this last scenario if I fake the requests again I can reach more than 10,000 req/sec. I assume that on a bigger VM could be much more. Unfortunately this is quite resource consuming.

Regards,
Milko

milkotodorov:
And something in addition that I've noticed is that one of the CPUs of the gorouter VM is utilized much more than the others (above 90%), while some of the other CPUs are less than 50% utilized.

milkotodorov:
Additionally I replaced my app with NGINX httpserver (using the staticfile-buildpack) and there is no difference with the requests the the gorouter can handle. So I think this is its maximums. I saw a research question on the pivotal about this, but cannot write into there, so I post it here.

shashwathi:
@milkotodorov 
Thanks for reporting this!

[...]

PR #122, opened 2016-03-17
Support disabling access logging to file and syslog

youngm:
PR for https://github.com/cloudfoundry/gorouter/issues/117

I didn't notice before that gorouter already supported disabling loggregator logging.  Anyway I'd be fine changing this to `syslog_access_logging_enabled` if so desired.

### DIFF
--- a/access_log/create_running_access_logger.go
+++ b/access_log/create_running_access_logger.go
@@ -7,20 +7,20 @@ import (
 	"github.com/cloudfoundry/gorouter/config"
 	"github.com/pivotal-golang/lager"
 
-	"os"
 	"io"
+	"os"
 )
 
 func CreateRunningAccessLogger(logger lager.Logger, config *config.Config) (AccessLogger, error) {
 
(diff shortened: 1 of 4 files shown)


shalako:
Hi @youngm 
What do you mean by "loggregator logging"?

As part of resolving #119 our current plan is to disable syslog streaming of access logs. Then we'll revisit optionally enabling streaming of access logs.

Our current thinking is to log access logs locally by default (no manifest property to turn this off), and add a manifest property to optionally stream access logs to syslog if `syslog_daemon_config` is also configured (even when `syslog_daemon_config:true` access logs would not be streamed by default). In #117 we discussed a property that controlled whether access logs were logged at all; we'll want to reconsider this.

youngm:
@shalako It sounds like the planned changes in #119 will work for me.  I vote we close this PR and #117 assuming #119 gets resolved the way it is currently planned.

shalako:
Closing, as streaming of access logs will be disabled by default in #119

PR #131, opened 2016-04-29
Add max length check on HTTP method

markstgodard:
Hey @shalako 

Encountered a situation where invalid requests were being sent to `gorouter` and we would like to NOT forward these on.

For example, right now if I sent a request line that has a large http method (or lots of garbage), this will be accepted by `gorouter` and forwarded on to the backend host.

```
FOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOO / HTTP/1.1
Host: dora.bosh-lite.com
```

Based on the RFC, we cannot limit to the http methods to the standard
(HEAD, GET, POST, PUT, DELETE, CONNECT, OPTIONS, TRACE) as there could be custom http methods.

In golang 1.6, the http server now validates the http method in the request line, however it just validates that it is a valid token / characters, but does not validate or limit the length.
Request line has no limits that I know of.

This PR is to put a sensible max limit length on the request line's http method (i.e. 64 chars) and if the method is greater than this max, the gorouter will return a 400 Bad Request (invalid http method)

```
HTTP/1.1 400 Bad Request
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
X-Vcap-Request-Id: 0abba867-865f-43a3-55de-787c02f5275f
Date: Fri, 29 Apr 2016 03:56:48 GMT
Content-Length: 37

400 Bad Request: invalid http method
```

I also made this configurable in case there are scenario's where operators would like to reduce or increase.  I also made sure that we cannot set this to lower than the max length of the standard http methods (i.e. CONNECT = length of 7).

I also have changes for cf-release to add to spec, templates, etc.

Please have a review and let me know.

Cheers

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -134,6 +134,8 @@ type Config struct {
 	TokenFetcherRetryIntervalInSeconds        int    `yaml:"token_fetcher_retry_interval"`
 	TokenFetcherExpirationBufferTimeInSeconds int64  `yaml:"token_fetcher_expiration_buffer_time"`
 
+	MaxHttpMethodLength int `yaml:"max_http_method_length"`
+
 	PidFile string `yaml:"pid_file"`
 }
 
@@ -159,6 +161,8 @@ var defaultConfig = Config{
 	TokenFetcherMaxRetries:                    3,
 	TokenFetcherRetryIntervalInSeconds:        5,
 	TokenFetcherExpirationBufferTimeInSeconds: 30,
+
+	MaxHttpMethodLength: 64,
 }
 
 func DefaultConfig() *Config {
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -265,6 +265,11 @@ token_fetcher_expiration_buffer_time: 40
 			Expect(config.TokenFetcherRetryIntervalInSeconds).To(Equal(5))
 			Expect(config.TokenFetcherExpirationBufferTimeInSeconds).To(Equal(int64(30)))
 		})
+
+		It("sets the default max http method length", func() {
+			Expect(config.MaxHttpMethodLength).To(Equal(64))
+		})
+
 	})
 
 	Describe("Process", func() {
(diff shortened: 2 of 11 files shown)


shalako:
This seems like an artificial limitation, no? Does the RFC say anything about a max? Has this caused a problem somewhere?

Why do we need a limit?

markstgodard:
Hey Shannon

This has caused a problem in that if you had, for example, a large number of invalid requests, we were seeing it consume a large amount of gorouter resources.
The reasons could range from a misbehaving application or malicious requests.

There is no predefined limit on the **request line**
```request-line   = method SP request-target SP HTTP-version CRLF```
https://tools.ietf.org/html/rfc7230#section-3.1.1 

Note: Most browsers put their own limit on **request line** size, and RFC wise, it seems like that is legit.. you just need respond accordingly.

I do not see a limit on *token* either, just limited to a set of valid chars.

In our case we were receiving invalid requests with the first part of the request (token) being hundreds of chars long and it was being forwarded on:

i.e.   [100s chars of garbage]GET /somepath HTTP/1.1

And in this case it would be fwd'd on as normal to the backend.

I also did investigation into newer version of golang, and in 1.6 they validate the method,  but they just check for valid chars.

In discussion with @fraenkel he suggested since we can't validate the content of method, we could have a max length check on *method*.

fraenkel:
@markstgodard Like I mentioned I would rather limit the request line but I think it should be something we should have golang deal with since it won't stop reading.
Checking the method once you have the entire request line is a bit odd since it is for your specific issue and not very generic.

markstgodard:
@fraenkel Agreed. Limiting the request line is ideal. I can open an enhancement request in golang and seeing if they are open to having a request line limit.
(I assume this would have to be optional, since the RFC indicates no limit)

@shalako I'll close this one

Cheers

PR #138, opened 2016-06-29
Configure HealthCheck User-Agent header

saliceti:
# What

Allow configuring the User-Agent header recognised as a healthcheck request. For example the Amazon ELBs send header `ELB-HealthChecker/1.0` in their healthchecks.
This is essential to take advantage of the drain feature of gorouter and be able to shutdown gracefully without interrupting any connection.
The original "HTTP-Monitor/1.1" is maintained as default.

# How to review
* Run gorouter and configure using:

```
drain_wait: 60
healthcheck_user_agent: test_header
```

* Configure a backend app as per https://github.com/cloudfoundry/gorouter#example
* `curl my_first_url.vcap.me:8081` (app) and `curl -A "test_header" "http://localhost:8081"` (healthcheck) should return 200
* Send signal USR1 to the process: `pkill -USR1 gorouter`. This starts draining process that will last 60s until gorouter stops.
* During this time:
  * `curl my_first_url.vcap.me:8081` should return 200
  * `curl -A "test_header" "http://localhost:8081"` should return 503 (to notify an upstream load balancer to stop sending new connections)

# Note

This should be followed by a PR on routing-release to allow using the new `healthcheck_user_agent` property.


proxy/proxy.go
```diff
@@ -421,8 +424,8 @@ func isProtocolSupported(request *http.Request) bool {
 	return request.ProtoMajor == 1 && (request.ProtoMinor == 0 || request.ProtoMinor == 1)
 }
 
-func isLoadBalancerHeartbeat(request *http.Request) bool {
-	return request.UserAgent() == "HTTP-Monitor/1.1"
+func isLoadBalancerHeartbeat(request *http.Request, healthCheckUserAgent string) bool {
```
keymon: This could be a private method of proxy struct. This way you can consume the `p.healthCheckUserAgent` directly, and you decouple the implementation details of `isLoadBalancerHeartbeat()` from the caller.

But that is a minor detail.

crhino:
@saliceti Thanks for the PR, we definitely want to support ELB HTTP health checks in some way.

In thinking about this feature, one question I have is whether it would be more useful to not expose a configuration property and instead add `ELB-HealthChecker/1.0` to a list of health check user agents.
My reasoning here is that 1) all operators of CF using ELBs would benefit without having to update their manifest and 2) would enable operators to switch their load balancers without having to remember to reconfigure the router with a different health_check_agent.

Of course, a downside of this idea would be maintaining this list of user agents and perhaps being too specific about ELB support. Let me know what you think of this, and I can talk to the other members of the routing team as well.

saliceti:
This is an interesting idea but I don't feel 100% comfortable with it.

I agree there would be a cost of maintaining a list of user agents. And some LB healthchecks are completely customisable, like F5. How would we handle this?

Another concern is performance. Gorouter already consumes a lot of CPU and I wonder about the additional cost of comparing this header against a list of known user agent strings, for each request. If the idea is to add more and more user agents, this might not be sustainable.

shalako:
Prioritizing for engineering review.

dcarley:
Thanks!

Issue #132, opened 2016-05-10
Gorouter likely to lose routes when NATS VM crashes hard


CAFxX:
We found out by hard crashing NATS VMs that gorouter (as well as other components, like CC) fail to quickly failover to the other NATSes in the cluster. This does not happen if the NATS process crashes (in this case the OS on the NATS VM will close the sockets), it happens only when the whole VM crashes.

While other components are also affected, gorouter is the worst affected because this problem leads to lost traffic.

All of our gorouter settings for timeouts are the default ones. We're using cf-release 220, both NATS and gorouter are on openstack, the stemcell in use is 3215.

In practice we see that roughly 100-110s after the crash of the NATS VM the routers that were connected to that specific NATS VM forget the routes. After a few minutes the disconnection is finally detected and a new connection established.

We're still investigating and confirming, but the most likely culprit is in the pingInterval setting: https://github.com/cloudfoundry/gorouter/blob/64cf29ce174e04a955b883e0d270e158d67bd176/config/config.go#L190-L203. While it's true that the NATS client will detect the disconnection after two failed ping attempts, it's not correct to say that the disconnection will happen within `2*PingInterval` (as the comment above that code says): it will actually happen _after_ 2 failed attempts, and this means that it will take between `2*PingInterval` and `3*PingInterval` to disconnect. (see https://github.com/nats-io/nats/blob/6f8a5734602782ce3ab3874b474a1b215acc7ed8/nats.go#L2186-L2193)

To be on the safe side, we should also consider that reconnection is not immediate (it should be fast, but it's not immediate) so my guess is that the full, worst-case reconnection time is `3*PingInterval + (NumNATSServers-1)*NATSConnectionTimeout` and all of this should be lower than `dropletstalethresholdinseconds/2 - startresponsedelayintervalinseconds`. This is the bare minimum, we may even want to add a small extra time buffer to avoid playing with fire.

Until this is fixed, a crash of one of the NATS VMs will most likely result in lost traffic.

markstgodard:
Hi @CAFxX 

We recently worked on a story related to gorouter and failing over to another NATS instance  and pruning routes:  https://www.pivotaltracker.com/story/show/110199022

This is not yet in an official CF release, but should be in the next.

Essentially gorouter was pruning routes BEFORE reconnecting to a new NATS instances. We also had to review and reconcile all the "intervals" and "thresholds" that we had going on that related to gorouter, deas, diego and NATS.  

We will review and see if we have your case covered.

[...]

PR #140, opened 2016-07-11
Optional suspend route pruning if NATS unavailable

markstgodard:
Currently, in the event that a NATS cluster goes down or becomes unavailable, gorouter will continue to prune routes. More importantly, gorouter will actually attempt to reconnect to all configured NATS servers and if exceeds the NATS client MaxReconnect attempts, it will exit and monit will keep restarting gorouter. In this catastrophic scenario, gorouter will not retain any routing info and apps will not be routable.

This approach favors consistency over availability, however if an operator would prefer to keep apps routable in the event of a NATS cluster outage, this PR provides that option.

The PR does the following:
- Opt-in config **suspend_pruning_if_nats_unavailable** to suspend route pruning if we cannot connect to NATS
- Set max reconnect in NATS client to -1 (no limit) when suspend prune enabled
- if pruning was suspended at any point and when NATS becomes available, the routes 'updated' timestamp will be marked with a current timestamp, so that it "resets the clock" for routes and avoids gorouter from pruning all the routes once NATS comes back online
- Default behavior will remain as pruning routes and not override max reconnect (current behavior)

_NOTE: This strategy favors availability over consistency and there is a possibility of routing to an incorrect endpoint in the case of port re-use. To be used with caution._

Once this PR is merged, I can update cf-release spec, templates to include **router.suspend_pruning_if_nats_unavailable** with the default value being **false**.

**Acceptance:**
- enable "suspend_pruning_if_nats_unavailable" setting to true (can change by hand in gorouter.yml or via manifest change)
- restart gorouter
- start watching the gorouter routing table
i.e.
```
while true; do clear && curl -s http://router:router@IP_ADDRESS_OF_GOROUTER:8080/routes | jq . && sleep 3; done
```
- bring down all NATS instances one by one
- gorouter.log should indicate "prune-suspended"
- wait for period of time until prune stale droplet threshold exceeded (or longer)
- routes should still be showing up in watch of routing table
- bring back nats instances
- routes should remain showing up in watch of routing table

### DIFF
--- a/registry/registry_test.go
+++ b/registry/registry_test.go
@@ -1,6 +1,9 @@
 package registry_test
 
 import (
+	"fmt"
+	"net"
+
 	"github.com/cloudfoundry-incubator/routing-api/models"
 	. "github.com/cloudfoundry/gorouter/registry"
 	. "github.com/onsi/ginkgo"
@@ -622,6 +625,7 @@ var _ = Describe("RouteRegistry", func() {
 				configObj.PruneStaleDropletsInterval = 50 * time.Millisecond
 				configObj.DropletStaleThreshold = 100 * time.Millisecond
 				reporter = new(fakes.FakeRouteRegistryReporter)
+
 				r = NewRouteRegistry(logger, configObj, reporter)
 			})
 
@@ -638,6 +642,51 @@ var _ = Describe("RouteRegistry", func() {
 				Expect(logger).ToNot(gbytes.Say(`prune.*"log_level":0.*foo.com/bar`))
 			})
 		})
+
+		Context("when suspend pruning is triggered (i.e. nats offline)", func() {
+			var totalRoutes int
+
+			BeforeEach(func() {
+				totalRoutes = 1000
+				Expect(r.NumUris()).To(Equal(0))
+				Expect(r.NumEndpoints()).To(Equal(0))
+
+				// add endpoints
+				for i := 0; i < totalRoutes; i++ {
+					e := route.NewEndpoint("12345", "192.168.1.1", uint16(1024+i), "id1", nil, -1, "", modTag)
+					r.Register(route.Uri(fmt.Sprintf("foo-%d", i)), e)
+				}
+
+				r.StartPruningCycle()
+				r.SuspendPruning(func() bool { return true })
+				time.Sleep(configObj.PruneStaleDropletsInterval + 10*time.Millisecond)
+			})
+
+			It("does not remove any routes", func() {
+				Expect(r.NumUris()).To(Equal(totalRoutes))
+				Expect(r.NumEndpoints()).To(Equal(totalRoutes))
+
+				time.Sleep(configObj.PruneStaleDropletsInterval + 50*time.Millisecond)
+
+				Expect(r.NumUris()).To(Equal(totalRoutes))
+				Expect(r.NumEndpoints()).To(Equal(totalRoutes))
+			})
+
+			Context("when suspend pruning is turned off (i.e. nats back online)", func() {
+				It("marks all routes as updated and does not remove routes", func() {
+					Expect(r.NumUris()).To(Equal(totalRoutes))
+					Expect(r.NumEndpoints()).To(Equal(totalRoutes))
+
+					r.SuspendPruning(func() bool { return false })
+
+					time.Sleep(configObj.PruneStaleDropletsInterval + 10*time.Millisecond)
+
+					Expect(r.NumUris()).To(Equal(totalRoutes))
+					Expect(r.NumEndpoints()).To(Equal(totalRoutes))
+				})
+			})
+		})
+
 	})
 
 	Context("Varz data", func() {
@@ -687,3 +736,13 @@ var _ = Describe("RouteRegistry", func() {
 		Expect(string(marshalled)).To(Equal(`{}`))
 	})
 })
+
+func portListening(addr string) bool {
+	var ok bool
+	conn, err := net.Dial("tcp", addr)
+	if err == nil {
+		ok = true
+		defer conn.Close()
+	}
+	return ok
+}
--- a/config/config.go
+++ b/config/config.go
@@ -99,13 +99,14 @@ type Config struct {
 	CipherString string `yaml:"cipher_suites"`
 	CipherSuites []uint16
 
-	PublishStartMessageIntervalInSeconds int `yaml:"publish_start_message_interval"`
-	PruneStaleDropletsIntervalInSeconds  int `yaml:"prune_stale_droplets_interval"`
-	DropletStaleThresholdInSeconds       int `yaml:"droplet_stale_threshold"`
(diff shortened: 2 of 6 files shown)


shalako:
Hi Mark,
Just back from vacation and reviewing this. 

- Set max reconnect in NATS client to -1 (no limit) when suspend prune enabled

what does max reconnect mean?

Thanks

markstgodard:
Max reconnect set to -1 in nats client means to always try reconnecting. Our current setup will abort after 10 attempts and then our callback handler in gorouter will exit. Then monit does its thing with gorouter process

shalako:
Do you mean that after 10 attempts gorouter effectively crashes and when monit restarts it, gorouter will try 10 more times? 

If this new property is set to true, and NATS is unavailable on a deploy, will the deploy fail if the routers are restarted? We like to give operators fast feedback while they are in the BOSH context. 

Maybe we should do the continuous retry all the time, rather than only when this property is set to true. But on a deploy, where the routers are bounced, if NATS is unavailable then the deploy should fail as a result of router failing to come up.

Is it possible to fail the deploy but leave the router in a running state, continuously attempting to contact NATS? I'm guessing in order to fail the deploy we have to report the process failing to monit.

markstgodard:
Yes, once we exceed the default nats MaxReconnect, we receive a Closed callback which gorouter has registered a func to log and exit. Monit then sees it as failed and does the restart dance. 

With MaxReconnect set to -1 (in the case we want to suspend pruning for nats) the connect to Nats logic on gorouter behaves the same way, we will fail. We have special logic on gorouter startup to try 3 times and if we did not connect successfully then we exit and monit does its thing.  

The MaxReconnect  option set to -1 comes into play when we have made at least a successful connect once at startup. 

So the fail fast behavior on startup is still in place. 

Cheers

shalako:
If `suspend_pruning_if_nats_unavailable:true`, then `MaxReconnect: -1`. In one comment above you say that this results in gorouter will continue retrying to connect to NATS forever. Then you say this config results in gorouter will fail, triggering monit to restart. I'm confused.

To confirm; the startup behavior is the same, no matter what `suspend_pruning_if_nats_unavailable` is set to?

markstgodard:
Correct. Startup behavior is the same no matter what suspend_pruning_if_nats_is_unavailable is set to. After successful startup, then the max reconnect -1 will have meaning and will not fail and exit after exceeds the default 10 retries. Sorry should have clarified that startup had same semantics 
Cheers

shalako:
> If suspend_pruning_if_nats_unavailable:true, then MaxReconnect: -1. In one comment above you say that this results in gorouter will continue retrying to connect to NATS forever. Then you say this config results in gorouter will fail, triggering monit to restart. I'm confused.

>  After successful startup, then the max reconnect -1 will have meaning and will not fail and exit after exceeds the default 10 retries.

Please clarify whether there is a difference in behavior with regard to gorouter exiting and being restarted by monit when `suspend_pruning_if_nats_is_unavailable` is true of false. Does gorouter exit in one case but not in another?

markstgodard:
Hi @shalako 

The `suspend_pruning_if_nats_is_unavailable` setting will NOT affect the gorouter initial startup behavior related to failing fast if gorouter cannot connect to NATS. 

On initial startup, if gorouter cannot connect to NATS server(s), gorouter will exit with a non-zero exit code  (regardless of the `suspend_pruning_if_nats_is_unavailable` setting **true** or **false**)

The `suspend_pruning_if_nats_is_unavailable` setting will only come into play once gorouter has started up successfully (meaning it has connected to NATS on startup).

If the gorouter has successfully started up and if all NATS server(s) go away for an extended period of time and if `suspend_pruning_if_nats_is_unavailable` = **true**, then gorouter will not prune routes  and will continue attempting to connect to NATS servers (nats client max reconnect -1 )

If the gorouter has successfully started up and if all NATS server(s) go away for an extended period of time and if `suspend_pruning_if_nats_is_unavailable` = **false**, then gorouter will exit when the nats client connection closes after exceeds max reconnects (nats client max reconnect set to nats client default [60](https://github.com/nats-io/nats/blob/master/nats.go#L34))


**TL;DR;**
- gorouter startup behavior of "fail fast" is the same
- gorouter will eventually exit if `suspend_pruning_if_nats_is_unavailable` = **false** and if NATS goes away for an extended period of time
- gorouter will NOT exit if `suspend_pruning_if_nats_is_unavailable` = **true** and if NATS goes away for an extended period of time

Feel free to ping me if you wanted to clarify anything

Cheers

shalako:
Thank you for the clarification. Prioritizing for teams' review.

registry/registry_test.go
```diff
@@ -687,3 +736,13 @@ var _ = Describe("RouteRegistry", func() {
 		Expect(string(marshalled)).To(Equal(`{}`))
 	})
 })
+
+func portListening(addr string) bool {
```
crhino: @markstgodard Seems like these function is not used anywhere?

markstgodard: Hey @crhino, good catch... it was a test helper function, I must have refactored the test and forgot to dd it.  I can nix it.

crhino:
@shalako Are you OK with the name `suspend_pruning_if_nats_is_unavailable`? We are thinking about the possibility of implementing the same feature for Routing API, and having to add another property like `suspend_pruning_if_routing_api_is_unavailable`.

@markstgodard Commented on the PR itself, seems like an unused test function snuck in, but other than that everything looks good.

markstgodard:
@shalako @crhino We could have a more generic name for the property. And then in subsequent work, could add routing-api as another source.

shalako:
I'm ambivalent about the name. I like that the current name is descriptive. If we used a generic name (also ok) the description would have to note that if the routing api is unavailable and NATS is available, routes will be pruned.

crhino:
@flawedmatrix and I discussed and decided we also like the descriptive nature of the current name. If we need to add another property later for routing-api that is not so bad.

jzlwang:
@markstgodard you mentioned that when `suspend_pruning_if_nats_is_unavailable` is true, we should see a "prune-suspended" message in `gorouter.log`. While doing acceptance I didn't see any logs at all in `gorouter.log` or `gorouter.err.log`. Can you please fix this for us? Thanks.

markstgodard:
@jzlwang Hey Judy, right I had that log statement as `debug` level, so the default logging level would not show that.  Did we what that log statement as `info` level?  Seems reasonable.

markstgodard:
Hey Judy, I don't push access anymore, so I had to submit a PR:
https://github.com/cloudfoundry/gorouter/pull/145
Also I didnt want to add commits to this specific PR since its closed.

Just get one of the team to review and merge if you want this to show up as info level.

Cheers

jzlwang:
Awesome, thanks Mark!

CAFxX:
@flawedmatrix @markstgodard I haven't tried yet the fixed version but from a check of this fix I'd say it won't solve the problem I reported in #132 (that was closed because #140 was merged)

Issue #141, opened 2016-07-13
GoRouter on Golang 1.5 will stop accepting connections when Proxy Protocol is enabled and a client connects without sending data


keymon:
Note: This issue comes from [this comment by @shashwathi](https://github.com/cloudfoundry/gorouter/pull/126#issuecomment-231470118)

# What

GoRouter recently added Proxy Protocol support by merging #126 . That PR uses the [go-proxyproto library](https://github.com/armon/go-proxyproto) which is affected by this bug https://github.com/armon/go-proxyproto/issues/1 when using Golang 1.5. 

GoRouter uses Golang 1.5 so it is affected.  When a client connects to gorouter with proxy protocol enabled but does not send any data, it will block gorouter and prevent it from accepting new connections.

## Impact

 * Go router does behave normally and accept multiple connections, until the issue is triggered.
 * This issue happens only when Proxy Protocol is enabled, it does not affect the original behaviour.
 * This issue happens only if the incoming connection doesn't actually send over any data.
    * when behind a Proxy which uses Proxy Protocol, it will always send the PROXY header.
    * normal HTTP clients will always send HTTP headers.

# How to reproduce

As described in [this comment by @shashwathi](https://github.com/cloudfoundry/gorouter/pull/126#issuecomment-231470118)

> Steps to reproduce this issue:
>  * Enable proxy protocol on gorouter
>  * Use telnet to establish a long lived connection to GOROUTER_IP:PORT **but not not send any data**
>  * You can use curl or telnet to establish another connection to GOROUTER_IP:PORT. If you are using curl you will notice your request just hangs.
>
> When you disable proxy protocol and repeat the above steps, you will see that the gorouter accepts multiple connections on its port.
>
> We have [a failing test case on a branch](https://github.com/cloudfoundry/gorouter/tree/proxy-proto) in gorouter to demonstrate the same. 

## Root cause

The root cause is described in  https://github.com/armon/go-proxyproto/issues/1

## Solution

Possible solutions are: 

 * Use Goland 1.6, which is not affected by the bug.
 * Patch go-proxyproto to handle this situation, for example by timing out if no data is sent over. This approach [has been implemented here](https://github.com/armon/go-proxyproto/pull/4)

keymon:
We implemented the second option *Patch go-proxyproto to handle this situation, for example by timing out if no data is sent over* in this PR https://github.com/armon/go-proxyproto/pull/4.

[...]

PR #143, opened 2016-07-19
Add go-proxyproto 100ms timeout for Proxy Protocol Header

keymon:
# What 

As described in #141, currently the Proxy Protocol implementation has a bug that might cause go router stop accepting connections when using Proxy Protocol and a client connects without sending data. 

Solutions are upgrade to Golang 1.6 used a patched version of upstream go-proxyproto  which allows define a timeout to receive the Proxy Header.

This PR configure a 100ms timeout to receive the Proxy Header Protocol. 

# Consequences 

This PR:
 * hardcodes the timeout 100ms
 * assumes that 100ms is enough time to receive the Proxy Header from the proxy. 
     * That is very likely the case, as the proxy will send the header immediately after stablish the connection, and hopefully there is not such latency between the proxy and the gorouter!!!
     * In the worse case scenario, if it timeouts after 100ms, the connection will be drop as invalid HTTP request
 * If a client connects and does not send data, gorouter will still be block for 100ms and not accepting connections. This might degrade the performance if there is are clients constantly connecting and not sending data.

# Dependencies and additional work

**IMPORTANT:** 

This code depends on https://github.com/armon/go-proxyproto/commit/3daa90aec0039a806299b9078f4422fee950f33c and **it is not compatible with previous versions of the library**

Once this PR is merged, https://github.com/cloudfoundry-incubator/routing-release must be updated  with the submodule references for github.com/cloudfoundry/gorouter and https://github.com/armon/go-proxyproto

I will not do a PR on routing-release as it needs to be updated by whoever merges this.

# How to test

We included the [test added](https://github.com/cloudfoundry/gorouter/pull/126#issuecomment-231470118) by @shashwathi to check that multiple connections can be established.

Thx

(our project story https://www.pivotaltracker.com/story/show/126483997)

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -39,7 +39,8 @@ import (
 var DrainTimeout = errors.New("router: Drain timeout")
 
 const (
-	emitInterval = 1 * time.Second
+	emitInterval               = 1 * time.Second
+	proxyProtocolHeaderTimeout = 100 * time.Millisecond
 )
 
 var noDeadline = time.Time{}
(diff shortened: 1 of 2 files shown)


keymon:
Build is failing because the dependencies must be updated in  https://github.com/cloudfoundry-incubator/routing-release.

shashwathi:
We've merged in the PR as part of this commit: https://github.com/cloudfoundry/gorouter/commit/45c41d9c556342bf27bf9b5f0b239cf618575a0e

PR #144, opened 2016-08-04
Track response metrics by component.

jmcarp:
Track response code counts by component, for parity with the `/varz` endpoint.

### DIFF
--- a/metrics/metrics_reporter.go
+++ b/metrics/metrics_reporter.go
@@ -40,7 +40,9 @@ func (m *MetricsReporter) CaptureRoutingRequest(b *route.Endpoint, req *http.Req
 }
 
 func (m *MetricsReporter) CaptureRoutingResponse(b *route.Endpoint, res *http.Response, t time.Time, d time.Duration) {
-	dropsondeMetrics.BatchIncrementCounter(getResponseCounterName(res))
+	responseCounterName := getResponseCounterName(res)
+
+	dropsondeMetrics.BatchIncrementCounter(responseCounterName)
 	dropsondeMetrics.BatchIncrementCounter("responses")
 
 	latency := float64(d / time.Millisecond)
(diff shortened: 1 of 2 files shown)


shalako:
Hi @jmcarp 

In the past, gorouter did a lot of counting and math for metrics exposed on `/varz`. Rather than add these things again to dropsonde, it seems to us that counting events by response code could be handled by an aggregator hooked up to the firehose. 

What do you think?

jmcarp:
I might be misunderstanding, but I thought that component logs didn't go through the firehose. Do you mean setting up a syslog drain and aggregating component responses by code there instead? 

Also, gorouter is already sending metrics on latency grouped by component, so I figured that grouping response codes by component would make sense too.

shalako:
Gorouter access logs do go through the firehose, in addition to HttpStartStop events. Messages with 5xx status codes could be counted for URIs of system components.

You mentioned the documentation found here: https://docs.pivotal.io/pivotalcf/1-7/opsguide/metrics.html#router-column

We'll ask the CloudOps team how they are generating their Router Dial Error metrics.

shashwathi:
@jmcarp : What is the motivation behind this PR? 

I would like to understand the need for tracking internal component response metrics. Are you seeing more failures with internal components? 

I think it is valuable to track response metrics for (internal)components separately from application instances. Does it solve your problem? @jmcarp 

@shalako @abbyachau thoughts? 
Regards
Shash

shashwathi:
@jmcarp : 
Looks like files in this PR got renamed so merging this might be more harder. Can you please update the PR, otherwise we might have to cherry pick the changes to gorouter.

Regards
Shash

shashwathi:
@jmcarp : Closing this issue due to inactivity. Feel free to open another PR if you still require this feature.

Regards
Shash

PR #147, opened 2016-08-25
Support for Least connection load balancing

markstgodard:
This PR will provide support for gorouter to use an alternate load balancing algorithm vs the default round-robin approach.

For more information, please see: https://docs.google.com/document/d/1EKzq7Bh0_7e8iwT8ikjlLMvq7prEyuBFysWCLSolBnY/edit?usp=sharing

This will be configured at the gorouter level, and could be changed by an operator in the deployment manifest (router.default_balancing_algorithm) which would update the gorouter.yml config.

**gorouter.yml**
```yaml
default_balancing_algorithm: least-connection
```

**Refactor:**
- refactored out the default round-robin algorithm
- added least-connection algorithm

**Support for least-connection load balancing algorithm**
- will select endpoint with the least number of connections
- if multiple endpoints match with the same number of least connections, will
  select a random one within those least connections
- currently scoped at gorouter level and can replace the default round-robin
- adds connection tracking stats

_NOTE: This currently only supports changing the load balancing strategy at the gorouter level and not at a finer-grained level such as route-level._  

# Testing:

## Manual functional tests
Outside of the unit tests, I did some tests using an app that I could control the response times:
https://github.com/markstgodard/sloth
Using this I did some tests where I could have one app slow than the other (i.e. 100ms, 2s, etc.) and compare and contrast the results between LC and RR.
Note, this was intentionally testing a scenario where one app was consistently laggier, however it illustrates that LC was working in that it was not straight RR-ing.

Please see this gist with more info: https://gist.github.com/markstgodard/8174efa110fdd824607b94d0ef441a48

## routing perf tests
I also used the https://github.com/cf-routing/routing-perf-release and was able to test this on bosh-lite by doings runs where I used RR vs LC.
I did 75 runs of the routing-perf release http_performance_tests errand.
30 times with RR, 45 times w/ LC.   Nothing really conclusive regarding if LC is slower than RR in the case where there is a single route and endpoint.  Sometimes slower, sometimes faster. Could just be related to bosh-lite.

I documented my results here: https://docs.google.com/spreadsheets/d/1WWh144qa8Cjt390vj5HtsM9sgb0tKjkF-n85xLnQP3k/edit?usp=sharing

Also I have NOT yet submitted a PR for the manifest changes to add `default_balancing_algorithm`. Between runs I manually modified the `gorouter.yml` to add the 
```yaml
default_balancing_algorithm: least-connection
```
Note, valid values are "least-connection" or "round-robin", or if absent defaults to round-robin.

NOTE: Since I refactored some stuff in `route` package the scripts/test failed in `go vet` since its running on a per file basis.  I talked to Shash in Slack about it.
I also submitted a separate PR to address this issue. So this PR will not pass the go vet tests as is. 
See other PR: https://github.com/cloudfoundry/gorouter/pull/146

Please feel free to ping me on Slack or a hangout if you need to discuss, since this is a larger PR.

Cheers

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,7 +226,23 @@ go tool pprof http://localhost:8080/debug/pprof/profile
 
 ## Load Balancing
 
-The GoRouter is, in simple terms, a reverse proxy that load balances between many backend instances. The implementation currently uses simple round-robin load balancing and will retry a request if the chosen backend does not accept the TCP connection.
+The GoRouter is, in simple terms, a reverse proxy that load balances between many backend instances. The default load balancing algorithm that GoRouter will use is a simple **round-robin** strategy. GoRouter will retry a request if the chosen backend does not accept the TCP connection.
+
+### Round-Robin
+Default load balancing algorithm that gorouter will use or may be explicity set in **gorouter.yml**
+```yaml
+default_balancing_algorithm: round-robin
+```
+
+### Least-Connection
+The GoRouter also supports least connection based routing and this can be enabled in **gorouter.yml**
+```yaml
+default_balancing_algorithm: least-connection
+```
+Least connection based load balancing will select the endpoint with the least number of connections. If multiple endpoints match with the same number of least connections, it will select a random one within those least connections.
+
+_NOTE: GoRouter currently only supports changing the load balancing strategy at the gorouter level and does not yet support a finer-grained level such as route-level. Therefore changing the load balancing algorithm from the default (round-robin) should be proceeded with caution._
+
 
 ## PROXY Protocol
 
--- a/config/config.go
+++ b/config/config.go
@@ -14,6 +14,11 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+const LOAD_BALANCE_RR string = "round-robin"
+const LOAD_BALANCE_LC string = "least-connection"
+
+var LoadBalancingStrategies = []string{LOAD_BALANCE_RR, LOAD_BALANCE_LC}
+
 type StatusConfig struct {
 	Port uint16 `yaml:"port"`
 	User string `yaml:"user"`
@@ -137,7 +142,8 @@ type Config struct {
 	TokenFetcherRetryInterval                 time.Duration `yaml:"token_fetcher_retry_interval"`
 	TokenFetcherExpirationBufferTimeInSeconds int64         `yaml:"token_fetcher_expiration_buffer_time"`
 
-	PidFile string `yaml:"pid_file"`
+	PidFile     string `yaml:"pid_file"`
+	LoadBalance string `yaml:"default_balancing_algorithm"`
 }
 
 var defaultConfig = Config{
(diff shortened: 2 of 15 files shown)


shalako:
Hi @markstgodard 

Thank you for this.

Why is the property in your PR `default_balancing_algorithm` and not `balancing_algorithm`? Are you future-proofing the ability for developers to eventually choose the algo by route?

Also, could you please run the performance benchmarks on a real env? IIRC, we never saw trustworthy numbers from load testing on boshlite

markstgodard:
Hi @shalako  

I ran perf tests against `v241` vs `v241` w/ the `least-connection` load balancing strategy and what I saw was a roughly 2% difference in results: https://docs.google.com/spreadsheets/d/1zCoJZ9QziE9CV3uKJACSxPVJKqZmkdIyFKQTxrOnsDk/edit?usp=sharing

The LC strategy is technically doing more “work” than straight RR with regards to keeping track of the count of connections for a route (and this counter does need to be thread safe, so that may be the reason in the difference)

Gist of the first 10 runs of 50,000 requests @ 100 conncurrent per RR and LC: https://gist.github.com/markstgodard/4aa73e9ca8f749c56025b2d9f4798c80

Also these tests were single app, with single route, below are some runs w/ multiple app instances

I did another 10 runs per, this time w/ 3 app instances and also hit the gorouter directly, bypass ELB: https://gist.github.com/markstgodard/9b29165e471482bee2f89668ad074ba7    In this case, LC is 3.4% slower. It might be additional cost of keeping count of connections for LC. Or it could be the VM as well.  but it could be something else, vm, etc..  
In some of the runs LC actually performed slightly better, but on average (when averaging 10 runs of 50k @ 100 concurrent) LC was slightly slower.

https://docs.google.com/spreadsheets/d/1zCoJZ9QziE9CV3uKJACSxPVJKqZmkdIyFKQTxrOnsDk/edit?usp=sharing

Cheers

shalako:
Tests appear to be failing, but I have prioritized the story for Engineering review.

shalako:
@markstgodard Could you please add to this PR documentation on the gorouter README for load balancing options, with brief descriptions of the two supported modes?

markstgodard:
Hi @shalako 

- Re: tests failing   Yeah that is due to the way `go vet` was being run. I discussed w/ Routing team to not run on a per file basis, but to run on the whole directory.   **This has been fixed with PR #146  I submitted**

- Sure, I'll add README updates

markstgodard:
Hi @shalako 

I rebased, re-ran the unit tests and also added README updates.  Feel free to comment on the wording or tweak the README wording post PR merge.

Cheers

markstgodard:
@shalako 
Also looks like the tests are failing in Travis due to an unrelated issue,  main CI on master and Travis failing in this PR for the same reason.  @crhino posted on Slack channel that they are fixing that unit test failure.

Cheers

shalako:
I'm going to close this as the PR was merged and the feature is included in [routing-release v140](https://github.com/cloudfoundry-incubator/routing-release/releases/tag/0.140.0)

Thank you, @markstgodard

PR #152, opened 2016-09-28
Fix for sticky session bug in gorouter

andreas-kupries:
A customer reported that session stickiness did not work when used
together with routing based on context paths.

On investigation this was found to be an issue in gorouter, with it
mishandling applications paths containing upper-case characters, for
example `app/UP`.

The context path is extracted from the incoming uri after the actual
uri was processed by `RouteKey()`, which translates the **entire** uri
to lower case, instead of only schema and host, the only parts which
are actually case-insensitive as per RFC 3986, Section 6.2.2.1.

This modified, all-lowercase context path is then returned to the
browser via the `VCAP_ID` cookie. As the path information in the
cookie (`app/up` for our example) does not match the actual path
(`app/UP`) the browser ignores the cookie, does not send it back, and
the session does not stick.

The new test added to `registry_test.go` will show the faulty
behaviour and fail before the fix is applied.

The fix shown in the PR keeps the route key information in a separate
variable instead of overwriting the incoming uri.

This change as is not only fixes the issue in question, but also
ensures that the used route key itself is not changed at all compared
to before the change. This avoids possible issues with unstated
assumptions of other parts of the system, i.e. expecting an
all-lowercase key.

If such an assumption does not exist then an alternate fix would be to
rewrite the `RouteKey()` method to only (properly) treat host and schema
as case-insensitive, while keeping the path as is. This will also
retain the untranslated path, and further modify the routekey.

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -75,14 +75,15 @@ func (r *RouteRegistry) Register(uri route.Uri, endpoint *route.Endpoint) {
 
 	r.Lock()
 
-	uri = uri.RouteKey()
+	routekey := uri.RouteKey()
 
-	pool := r.byUri.Find(uri)
+	pool := r.byUri.Find(routekey)
 	if pool == nil {
 		contextPath := parseContextPath(uri)
 		pool = route.NewPool(r.dropletStaleThreshold/4, contextPath)
-		r.byUri.Insert(uri, pool)
-		r.logger.Debug("uri-added", lager.Data{"uri": uri})
+		r.byUri.Insert(routekey, pool)
+		r.logger.Debug("uri-added", lager.Data{
+			"uri": uri, "routekey": routekey})
 	}
 
 	endpointAdded := pool.Put(endpoint)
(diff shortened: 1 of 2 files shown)


andreas-kupries:
Looked over the CI log ...
```
registry/registry_test.go:144: not enough arguments in call to route.NewEndpoint
```
Hm ... I suspect my new check is using some older signature here, and I missed it as it wasn't a conflict.
Will fix.

andreas-kupries:
Issue reported by Travis should be addressed by commit 5fe5cc7f64f12a891fe20506a2f2d4de25d928b5.
Another issue found in separate review addressed in commit d23c8332e3c748df8a6c636fcf8091d5d0c7fb10, with existing test modified to capture this.

shashwathi:
@andreas-kupries : 

Thank you for bringing this to our attention. It does sound like there is a bug in gorouter. 
@shalako @abbyachau will prioritize the story.

Regards
Shash

abbyachau:
This has been merged and verified to work. The fix will be included in the next releases of `routing-release` and `cf-release`

Acceptance steps: https://gist.github.com/abbyachau/8351f51e4579ff8d3320721f0e893420

PR #161, opened 2017-01-20
Add tags as part of the return JSON of routes endpoint

samof76:
The foundation behind this change is that I am trying to use gorouter independently for dynamic route registration, and I would tags to identity difference kinds of services, to classify them based on tags and put them into various buckets.

### DIFF
--- a/route/pool.go
+++ b/route/pool.go
@@ -290,14 +290,16 @@ func (e *endpointElem) failed() {
 
 func (e *Endpoint) MarshalJSON() ([]byte, error) {
 	var jsonObj struct {
-		Address         string `json:"address"`
-		TTL             int    `json:"ttl"`
-		RouteServiceUrl string `json:"route_service_url,omitempty"`
+		Address         string            `json:"address"`
+		TTL             int               `json:"ttl"`
+		RouteServiceUrl string            `json:"route_service_url,omitempty"`
+		Tags            map[string]string `json:"tags"`
 	}
 
 	jsonObj.Address = e.addr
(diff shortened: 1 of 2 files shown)


shashwathi:
@samof76 : Thank you for submitting the PR. 

I am not able to understand how does updating `/routes` endpoint solve the problem of bucketing routes. This endpoint is intended for debugging purpose.

In PR description you have mentioned using `tags` to bucket services endpoints, do you feel that adding `tags` information is useful for debugging purpose on `/routes` endpoint? 

FYI: Please sign the appropriate CLA.
Regards
Shash

samof76:
@shashwathi Thanks for asking.

See, I am using go router as a HTTP router where a service registers itself in my case [netdata](http://my-netdata.io/), which monitors my machines on the cloud. These machine might be of different types, web frontends, web backends, middleware, databases, caches etc. I would want to build a dashboard that will help me search by these categories(of type), so it very helpful to have tags server as part of the `/routes` request.

So the model is like this, I have cron that every minute does a nats-pub from the instance where netdata is running, registering its netdata endpoint, along with tags say, `{"type": "memcached", "stack":"production", ...}`. No I would have an intermediate dashboard that reads the `/routes` and registers these classifications, and provide the user with the ability to filter based on them.

I hope I am clear.

shashwathi:
@samof76: 

Thank you for explaining.  I will ask @shalako @abbyachau to prioritize the story.

Regards
Shash

shashwathi:
Hi @samof76 : 

Fixed the PR and added more test coverage to explicitly test `tags`. Thank you for submitting the PR. 

Regards
Shash

PR #172, opened 2017-05-05
Improve performance of registry.Register

flawedmatrix:
The use of IncrementCounter vs BatchIncrement counter causes
registry.Registry to be 6~7x slower

with sender.IncrementCounter:
```
go test -run Benchmark -bench=.
BenchmarkRegisterWith100KRoutes-4         100000             23244 ns/op
BenchmarkRegisterWithOneRoute-4           100000             19677 ns/op
PASS
ok      code.cloudfoundry.org/gorouter/registry 17.593s
```

with batcher.BatchIncrementCounter:
```
go test -run Benchmark -bench=.
BenchmarkRegisterWith100KRoutes-4         300000              3616 ns/op
BenchmarkRegisterWithOneRoute-4           500000              3032 ns/op
PASS
ok      code.cloudfoundry.org/gorouter/registry 8.872s
```

This commit changes the behavior of one of Gorouter's metrics:
Instead of the `registry_message` counter metric being emitted on every
NATS message, it will emit the metric on a set interval.

### DIFF
--- a/metrics/metricsreporter.go
+++ b/metrics/metricsreporter.go
@@ -84,7 +84,7 @@ func (m *MetricsReporter) CaptureRegistryMessage(msg ComponentTagged) {
 	} else {
 		componentName = "registry_message." + msg.Component()
 	}
-	m.sender.IncrementCounter(componentName)
+	m.batcher.BatchIncrementCounter(componentName)
 }
 
 func (m *MetricsReporter) CaptureUnregistryMessage(msg ComponentTagged) {
--- a/metrics/metricsreporter_test.go
+++ b/metrics/metricsreporter_test.go
@@ -313,8 +313,8 @@ var _ = Describe("MetricsReporter", func() {
 			endpoint.Tags = map[string]string{}
 			metricReporter.CaptureRegistryMessage(endpoint)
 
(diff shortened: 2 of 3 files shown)


shalako:
@flawedmatrix Thank you for helping the team improve performance. 

Will the metric name change? 

How often will it be emitted? Based on other gorouter metrics documented [here](https://docs.cloudfoundry.org/running/all_metrics.html#gorouter), I would expect every 5 seconds.

On merging this PR, we will also need a PR to update the CF docs: https://github.com/cloudfoundry/docs-running-cf/blob/master/metrics/_routing.html.md.erb

flawedmatrix:
@shalako

The metric name isn't changed as part of this PR. It's emitted on the same interval as all the other batch metrics in gorouter (5 seconds).

Here's a performance report showing the comparison between how gorouter currently handles 100K routes and after making this improvement + the improvement in this PR: https://github.com/cloudfoundry/gorouter/pull/174

[registry_perf.pdf](https://github.com/cloudfoundry/gorouter/files/1003783/registry_perf.pdf)

shalako:
prioritized for eng team review

Issue #197, opened 2017-11-10
Instance routing should include in responses max(instance_idx)


CAFxX:
Thanks for submitting an issue to `gorouter`. We are always trying to improve! To help us, please fill out the following template.

## Issue

When using [app instance routing](https://docs.cloudfoundry.org/concepts/http-routing.html#app-instance-routing), I would like to see in the response headers the biggest instance_idx currently in the routing table, something like:

```
X-CF-APP-INSTANCE-MAX: 8
```
to signify e.g. that the maximum instance index currently running is 8.

## Context

When using instance routing for the purpose of monitoring individual instances, it would be very helpful if the routers returned the number of instances of the application. This would allow active monitoring systems to opportunistically discover the number of running instances without having to poll the CC API.

Ideally, this number should reflect the number of *desired* instances, not the number of *running* instances. Since it may be difficult for gorouter to obtain the *desired* number of instances of an application (BBS should forward this information to gorouter somehow) this can possibly be approximated with the highest currently running instance index (monitoring systems can then apply some form of hysteresis to the values returned)

## Steps to Reproduce

Send a request with the header `X-CF-APP-INSTANCE: $APP_GUID:$APP_INSTANCE`

## Expected result

I receive in the response `X-CF-APP-INSTANCE-MAX: $APP_INSTANCE_MAX`, where `$APP_INSTANCE_MAX` is the instance index of the currently-running instance with the greatest instance index.

## Current result

No such header is returned

## Possible Fix

When a request that makes use of app instance routing is received by the gorouter, gorouter should scan the instance list for the specified app and return the index of the instance with the greatest instance index.

shalako:
Seems reasonable. Would you send us a PR?

CAFxX:
@shalako sent an initial PR, any feedback is welcome... as soon as the implementation is agreed upon, I will add some more internal tests and run the ATs

KauzClay:
We are closing this issue due to the results of this [PR](https://github.com/cloudfoundry/gorouter/pull/198)

PR #198, opened 2017-11-13
Return the current maximum number of instances in X-CF-APP-INSTANCE-MAX

CAFxX:
Fixes https://github.com/cloudfoundry/gorouter/issues/197

To test, send a request using app instance routing and check that the response contains the `X-CF-APP-INSTANCE-MAX` header. The header is added as long as the `X-CF-APP-INSTANCE` request header is syntactically correct and it contains a valid app GUID.

Note: the implementation is less than elegant since it adds a field to the Pool to avoid changing interfaces and function signatures all over the place. Suggestions on how to improve it are welcome.

### DIFF
--- a/common/http/headers.go
+++ b/common/http/headers.go
@@ -9,6 +9,7 @@ const (
 	VcapTraceHeader       = "X-Vcap-Trace"
 	CfInstanceIdHeader    = "X-CF-InstanceID"
 	CfAppInstance         = "X-CF-APP-INSTANCE"
+	CfAppInstanceMax      = "X-CF-APP-INSTANCE-MAX"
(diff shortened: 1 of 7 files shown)


shalako:
@CAFxX 
My apologies for our delay in replying to this. Thank you very much for taking the time to submit this change.  I know you did so in part because I asked if you would.

Our team is wondering if this is still a problem for you, and why polling CC for this information is not an option. As you know, we're quite focused on the Istio integration and are trying to limit effort spent on the current routing subsystem to bug fixes.

Best,
Shannon

shubhaat:
@CAFxX we will revisit this when implementing instance based routing using Envoy. For now we are going to close this issue.

Issue #209, opened 2018-02-22
Keep Alive connections prematurely terminated by backend timeout code


univ0298:
## Issue

Keep alive connections are being closed because the backend timeout limit is not paying attention to whether a keep alive connection is alive because of ongoing re-use, or alive because of a long duration connection to a backend.

## Context

Keep alive will allow a connection to stay alive indefinitely, as long as it's used once every 90 seconds. 
However another feature, backend timeout maximum, is designed to cut off connections where a backend just takes too long to reply.
Unfortunately this second feature is colliding with keep alive, because the second feature sees a keep-alive connection as the same as a slow backend, and kills it, even if it is is mid-way through processing transactions and should have been kept alive.

This has a bad implication because it means that all keep alive transactions are terminated every n seconds, even if they are processing traffic. This leads to lots of 502 responses from the gorouter that should have been success responses.

## Steps to Reproduce

To reproduce this, be sure to use POST or PATCH requests. A GET request will not see the issue because the retry behavior "saves" the transaction and masks the issue. But POST/PATCH are not retry-able so will see it.
Also, a slow backend application makes it easier to recreate because the issue is seen by having an in-flight transaction get interrupted. So a simple ten second sleep in the app is helpful.

With that in place, it can be easily reproduced by configuring a request_timeout_in_seconds value of 30 seconds, and a high number of max_idle_connections (say 49000). Once this is done, send a POST request to an app. That will cause a connection from gorouter to app to be created and the app will reply (after 10 seconds if you have a sleep 10). Then 25 seconds after the first request was sent, send another one. You will see a 502 response code after 5 seconds (as the total test time hits 30 seconds).

Some more details of the scenario:
1.    A new connection established
2.    I/O Deadline set to Now()+endpoint_timeout (e.g. Now + 30s)  [see https://github.com/cloudfoundry/gorouter/blob/master/proxy/proxy.go#L131]
3.    Keep alive remains enabled for this connection
4.    Many requests occur over time on the Keep alive connection, still keeping it open
5.    A POST/PATCH request occurs on the kept alive connection close to the initial I/O deadline timeout set by (2)
6.    Before the response can be read from the request, the I/O deadline timeout expires and the connection is closed
7.    Since it was an outstanding POST/PATCH, the request is not retried and a 502 is returned to the caller

## Expected result

A keep alive connection should live indefinitely, as long as it's re-used within 90 seconds. No 502 error should be seen

## Current result

As explained above, transactions are failing.

[...]

PR #211, opened 2018-03-10
Set endpoint timeout on request context

swetharepakula:
- stop setting deadline at connection creation

Signed-off-by: Jonathan Berkhahn <jaberkha@us.ibm.com>

Set endpoint timeout per request, not per connection and set it on the context.

By setting the timeout on the connection, we cause keep-alive connections to be killed at the end of the timeout regardless of whether the connection is being actively used. 

BOSH Manifest:
Set `request_timeout_in_seconds` on the router vm to 30. The default is 900 seconds, we are reducing the timeout to show the error quicker.
Set `router.max_idle_connections` to a non zero number. We need just need to make keep-alive connections are alive.

Push up an app that has a deley on a POST endpoint. I added the following the default dora app:
```
post '/' do
    sleep 5
    "Hi, I'm Dora!"
  end
```

Then in a terminal you can run:
```SC=200; while [[ $SC -eq 200 ]]; do SC=`curl -X POST -s -o /dev/null -w "%{http_code}" dora.bosh-lite.com`; echo $SC; done```

You should see 200 for requests continuously

You should see 200 for request for about 30 seconds and then after 30 seconds you will see a 502.

Link to Issue this is solving: https://github.com/cloudfoundry/gorouter/issues/209

### DIFF
--- a/proxy/round_tripper/proxy_round_tripper.go
+++ b/proxy/round_tripper/proxy_round_tripper.go
@@ -1,6 +1,7 @@
 package round_tripper
 
 import (
+	"context"
 	"errors"
 	"io/ioutil"
 	"net/http"
@@ -68,6 +69,7 @@ func NewProxyRoundTripper(
 	localPort uint16,
 	errorHandler errorHandler,
 	routeServicesClient http.RoundTripper,
+	endpointTimeout time.Duration,
 ) ProxyRoundTripper {
 	return &roundTripper{
 		logger:              logger,
@@ -79,6 +81,7 @@ func NewProxyRoundTripper(
 		retryableClassifier: retryableClassifier,
 		errorHandler:        errorHandler,
 		routeServicesClient: routeServicesClient,
+		endpointTimeout:     endpointTimeout,
 	}
 }
 
@@ -92,6 +95,7 @@ type roundTripper struct {
 	retryableClassifier fails.Classifier
 	errorHandler        errorHandler
 	routeServicesClient http.RoundTripper
+	endpointTimeout     time.Duration
 }
 
 func (rt *roundTripper) RoundTrip(request *http.Request) (*http.Response, error) {
@@ -166,8 +170,7 @@ func (rt *roundTripper) RoundTrip(request *http.Request) (*http.Response, error)
 				// request from the gorouter will go to a backend using TLS (if tls_port is set on that endpoint)
 				tr = rt.routeServicesClient
 			}
-
-			res, err = tr.RoundTrip(request)
+			res, err = rt.timedRoundTrip(tr, request)
 			if err == nil {
 				if res != nil && (res.StatusCode < 200 || res.StatusCode >= 300) {
 					logger.Info(
@@ -233,11 +236,37 @@ func (rt *roundTripper) backendRoundTrip(
 
 	rt.combinedReporter.CaptureRoutingRequest(endpoint)
 	tr := GetRoundTripper(endpoint, rt.roundTripperFactory)
-	res, err := tr.RoundTrip(request)
+	res, err := rt.timedRoundTrip(tr, request)
 
 	// decrement connection stats
 	iter.PostRequest(endpoint)
 	return res, err
+}
+
+func (rt *roundTripper) timedRoundTrip(tr http.RoundTripper, request *http.Request) (*http.Response, error) {
+
+	if rt.endpointTimeout <= 0 {
+		return tr.RoundTrip(request)
+	}
+	reqCtx, cancel := context.WithTimeout(request.Context(), rt.endpointTimeout)
+	request = request.WithContext(reqCtx)
+	errored := make(chan struct{}, 1)
+
+	go func() {
+		defer cancel()
+		select {
+		case <-reqCtx.Done():
+		case <-errored:
+		}
+	}()
+
+	resp, err := tr.RoundTrip(request)
+	if err != nil {
+		cancel()
+		errored <- struct{}{}
+	}
+
+	return resp, err
 }
 
 func (rt *roundTripper) selectEndpoint(iter route.EndpointIterator, request *http.Request) (*route.Endpoint, error) {
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -50,6 +50,7 @@ type proxy struct {
 	forceForwardedProtoHttps bool
 	defaultLoadBalance       string
 	endpointDialTimeout      time.Duration
+	endpointTimeout          time.Duration
 	bufferPool               httputil.BufferPool
 	backendTLSConfig         *tls.Config
 }
@@ -118,21 +119,13 @@ func NewProxy(
 		forceForwardedProtoHttps: c.ForceForwardedProtoHttps,
 		defaultLoadBalance:       c.LoadBalance,
 		endpointDialTimeout:      c.EndpointDialTimeout,
+		endpointTimeout:          c.EndpointTimeout,
 		bufferPool:               NewBufferPool(),
 		backendTLSConfig:         tlsConfig,
 	}
 
 	httpTransportTemplate := &http.Transport{
-		Dial: func(network, addr string) (net.Conn, error) {
-			conn, err := net.DialTimeout(network, addr, p.endpointDialTimeout)
-			if err != nil {
-				return conn, err
-			}
-			if c.EndpointTimeout > 0 {
-				err = conn.SetDeadline(time.Now().Add(c.EndpointTimeout))
-			}
-			return conn, err
-		},
+		Dial:                (&net.Dialer{Timeout: c.EndpointDialTimeout}).Dial,
 		DisableKeepAlives:   c.DisableKeepAlives,
 		MaxIdleConns:        c.MaxIdleConns,
 		IdleConnTimeout:     90 * time.Second, // setting the value to golang default transport
@@ -152,6 +145,7 @@ func NewProxy(
 			ErrorSpecs:     round_tripper.DefaultErrorSpecs,
 		},
 		routeServicesClient,
+		p.endpointTimeout,
 	)
 
 	rproxy := &httputil.ReverseProxy{
--- a/proxy/round_tripper/proxy_round_tripper_test.go
+++ b/proxy/round_tripper/proxy_round_tripper_test.go
@@ -50,83 +50,90 @@ func (f *FakeRoundTripperFactory) New(expectedServerName string) round_tripper.P
 }
 
 var _ = Describe("ProxyRoundTripper", func() {
-	Context("RoundTrip", func() {
-		var (
-			proxyRoundTripper   round_tripper.ProxyRoundTripper
-			routePool           *route.Pool
-			transport           *roundtripperfakes.FakeProxyRoundTripper
-			logger              *test_util.TestZapLogger
-			req                 *http.Request
-			reqBody             *testBody
-			resp                *httptest.ResponseRecorder
-			alr                 *schema.AccessLogRecord
-			routerIP            string
-			combinedReporter    *fakes.FakeCombinedReporter
-			roundTripperFactory *FakeRoundTripperFactory
-			routeServicesClient *sharedfakes.RoundTripper
-			retryableClassifier *errorClassifierFakes.Classifier
-			errorHandler        *roundtripperfakes.ErrorHandler
+	var (
+		proxyRoundTripper   round_tripper.ProxyRoundTripper
+		routePool           *route.Pool
+		transport           *roundtripperfakes.FakeProxyRoundTripper
+		logger              *test_util.TestZapLogger
+		req                 *http.Request
+		reqBody             *testBody
+		resp                *httptest.ResponseRecorder
+		alr                 *schema.AccessLogRecord
+		routerIP            string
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

**Comment by shalako** [comment]
Thank you @swetharepakula and @jberkhahn ! I'm prioritizing this for the team to review. I hope you both are well.

**Comment by zachgersh** [comment]
I'd love to understand why we had a timeout set on the connection previously. Was this trying to achieve the same goal but we misunderstood where to put the timeout?

**Comment by shalako** [comment]
The only reason I can think of for a timeout on the connection to backend is in support of keepalives. I think our keepalive idle timeout to backends is 90s? We probably never caught this because the default for `request_timeout_in_seconds` is 900s. 

When `max_idle_connections: 0`, gorouter closes the tcp connection to the backend after the first http response.

**Comment by univ0298** [comment]
@shalako if I understand you correctly here .... the backend timeout is to ensure that a connection doesn't stay open indefinitely (or for longer than is desirable) in case of a slow backend application. In the case of no keep alive, that timeout makes sense and is simple. In the case of keep alive it's more complex because every time the connection is re-used, a fresh timer should be initiated on the connection, because it's the lifetime of the request that should be measured, not the lifetime of the connection (in the keep alive scenario these become separate).

So as I understand it, this problem occurred because KA was introduced as a function after the original request-with-backend-timeout function was in place. And the KA code appears to have overlooked that this backend timer should get reset to zero each time the connection is re-used. It's also important that the backend timeout is not lower than the 90 seconds keep alive time too.

File: proxy/round_tripper/proxy_round_tripper.go [hunk 1]
```diff
@@ -233,13 +236,39 @@ func (rt *roundTripper) backendRoundTrip(
...
+func (rt *roundTripper) timedRoundTrip(tr http.RoundTripper, request *http.Request) (*http.Response, error) {
+
+	if rt.endpointTimeout <= 0 {
+		return tr.RoundTrip(request)
+	}
+	reqCtx, cancel := context.WithTimeout(request.Context(), rt.endpointTimeout)
+	request = request.WithContext(reqCtx)
+	errored := make(chan struct{}, 1)
+
+	go func() {
```
zachgersh: @swetharepakula could you explain the purpose of this `goroutine` and the `select`? It currently appears unnecessary.

proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -233,13 +236,39 @@ func (rt *roundTripper) backendRoundTrip(
...
+		select {
+		case <-reqCtx.Done():
+		case <-errored:
+		}
+	}()
+
+	resp, err := tr.RoundTrip(request)
+	if err != nil {
+		cancel()
+		errored <- struct{}{}
```
zachgersh: Shouldn't we just be cancelling here and then returning there error (with an empty response). Function and its caller are already both setup that way.

proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -233,13 +236,39 @@ func (rt *roundTripper) backendRoundTrip(
...
 }
 
+func (rt *roundTripper) timedRoundTrip(tr http.RoundTripper, request *http.Request) (*http.Response, error) {
+
+	if rt.endpointTimeout <= 0 {
+		return tr.RoundTrip(request)
+	}
+	reqCtx, cancel := context.WithTimeout(request.Context(), rt.endpointTimeout)
+	request = request.WithContext(reqCtx)
+	errored := make(chan struct{}, 1)
```
zachgersh: I don't believe this `errored` channel is going to be necessary.

univ0298:
I think I see your point that the timeout should never have been on the connection though, only on the request

proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -233,13 +236,39 @@ func (rt *roundTripper) backendRoundTrip(
...
 	iter.PostRequest(endpoint)
 	return res, err
 }
 
+func (rt *roundTripper) timedRoundTrip(tr http.RoundTripper, request *http.Request) (*http.Response, error) {
+
+	if rt.endpointTimeout <= 0 {
+		return tr.RoundTrip(request)
+	}
+	reqCtx, cancel := context.WithTimeout(request.Context(), rt.endpointTimeout)
```
zachgersh: This bit is interesting due to our use of `reverse-proxy`. The go stdlib has already put a cancellation context onto the request that has been passed into this function. You'll see that cancellation get wired up here:

https://github.com/golang/go/blob/master/src/net/http/httputil/reverseproxy.go#L144-L165

That would potentially mean that `WithTimeout` could potentially instead be a `time.After` with the specified `rt.endpointTimeout` which would close the context `Done` chan if the timeout was reached. I think that may potentially work but the same but have not tried it.

proxy/round_tripper/proxy_round_tripper.go
swetharepakula: @zachgersh, the need for the goroutine is for the `defer cancel()`. We need to make sure that cancel gets called in all situations.

proxy/round_tripper/proxy_round_tripper.go
swetharepakula: @zachgersh The problem is that the gorouter relays most of the work about handling the request to the reverse proxy. The reason we have to set the `WithTimeout` is because only the reverse proxy knows when the request is finished. So even if we do a `time.After`, there is no capability for us to be able to know whether the request is finished to be able to cancel the request or not.

proxy/round_tripper/proxy_round_tripper.go
swetharepakula: We need the errored channel to be able to stop the go routine in the situation that we have canceled the context. In that situation we will be waiting on the channel from `ctx.Done()` indefinitely meaning we would start leaking goroutines.

proxy/round_tripper/proxy_round_tripper.go
swetharepakula: Sure we could return the resp, and err but it is the next line of code and we would have to return the same thing in all situations. The reason we need to check the error is to call cancel() and stop the goroutine. I do see that instead of calling cancel here we can expect the goroutine to do that instead so here we only send a signal on the error channel.

zachgersh:
This was manually merged.

shalako:
This appears to be working as expected:
1. Deployed release candidate with manifest properties:

  ```
  jobs:
  - name: gorouter
    properties:
      request_timeout_in_seconds: 30
      router:
        max_idle_connections: 49000
  ```

2. Curled dora app on existing `/delay` endpoint

  ```
  $ SC=200; while [[ $SC -eq 200 ]]; do SC=`curl -s -o /dev/null -w "%{http_code}" dora.superman.routing.cf-app.com/delay/5`; echo $SC; done
  ```

  Response is always 200

univ0298:
The test above is insufficient and would not have seen the original incident @shalako

univ0298:
Critically the test requires use of a POST request (see original details)

shalako:
Hi @univ0298,
I'm not sure why a POST is required. Gorouter doesn't know about HTTP methods. I saw that Swetha added a `POST /` endpoint for the sleep behavior. I don't know why that was necessary because the dora app already supports a sleep endpoint on `GET /delay/:seconds`. We can test with Swetha's  modification but I don't understand why it would make a difference.

See endpoint 9 at:
https://github.com/cloudfoundry/cf-acceptance-tests/blob/master/assets/dora/README.md

univ0298:
Hi @shalako . 

When I opened the original issue, https://github.com/cloudfoundry/gorouter/issues/209 , I tried to explain this:

> To reproduce this, be sure to use POST or PATCH requests. A GET request will not see the issue because the retry behavior "saves" the transaction and masks the issue. But POST/PATCH are not retry-able so will see it.

I don't know this to be 100% definitely correct but I am fairly certain that's why GET requests are not impacted.

shalako:
Thanks for the quick reply. We'll validate the expected behavior with Swetha's modification.

shubhaat:
Tested again. This seems to be working as expected with a POST too 

1. Set the manifest properties 

```
- name: gorouter
    properties:
      router:
        request_timeout_in_seconds: 30
        max_idle_connections: 10
```

2. Sent a request using 

`$ SC=200; while [[ $SC -eq 200 ]]; do SC=`curl -X POST -s -o /dev/null -w "%{http_code}" -H "Content-Length: 0"  dora.superman.routing.cf-app.com`; echo $SC; done`

Response: I get only 200 Responses. 

We will cut a release soon and let you know.

univ0298:
@shubhaat note that testing with a POST alone will not surface this issue. You need to be using a POST that is in-flight when the backend timeout expires. Therefore you need a POST endpoint that has a delay, like dora's GET /delay/:seconds . A simple POST with a fast reply, that it seems you tested with, will not suffice.

@shalako pls note

shubhaat:
@univ0298 - point noted. I actually did introduce a sleep in my app for the POST endpoint taking cue from the steps @swetharepakula mentioned in her initial description. So my POST looks like 

```
post '/' do
    sleep 5
    "Hi, I'm Dora!"
  end
```

shubhaat:
This is in routing-release 0.175.0

proxy/round_tripper/proxy_round_tripper.go
judwhite: @zachgersh @swetharepakula I think this code should look something like this:

```go
reqCtx, cancel := context.WithTimeout(request.Context(), rt.endpointTimeout)
defer cancel()
```

There doesn't appear to be a need for the goroutine or `errored` channel, and the `if err != nil` block can be simplified.

`RoundTrip` already handles `ctx.Err()` and returns it, so I don't think there's a need to check this.

proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -92,6 +95,7 @@ type roundTripper struct {
 	retryableClassifier fails.Classifier
 	errorHandler        errorHandler
 	routeServicesClient http.RoundTripper
+	endpointTimeout     time.Duration
 }
 
 func (rt *roundTripper) RoundTrip(request *http.Request) (*http.Response, error) {
```
judwhite: @swetharepakula @zachgersh This implementation of `http.RoundTripper` looks like it may violate some of the assumptions stated in godoc https://golang.org/pkg/net/http/#RoundTripper, in particular but possibly not limited to:

> RoundTrip should not modify the request, except for consuming and closing the Request's Body

I'll take a look at the overall intent. If there's anything to change I'll make a proposal.

proxy/round_tripper/proxy_round_tripper.go
swetharepakula: The need for the goroutine is because the function will exit before the complete response is returned. `tr.RoundTrip` returns as soon as the headers start coming in for the response, but it does not mean the response is completed. So if `defer cancel()` is added then the request is canceled as soon as the headers start coming in.
Here is the discussion about this particular PR: https://cloudfoundry.slack.com/archives/C05586PBX/p1520911938000018?thread_ts=1520893756.000433&cid=C05586PBX

judwhite: @swetharepakula @zachgersh Looking at this some more, I don't think `cancel()` ever gets called until the timeout elapses unless there's an error from `tr.RoundTrip`. While you're not leaking contexts they certainly hang around a while.

It looks like the context is created too late in the request pipeline, that's why you had to jump through hoops to get this to work. You should be able to call `cancel()` as soon as you're done with a context, not after the timeout fires.

proxy/round_tripper/proxy_round_tripper.go
judwhite: Just below what I can comment on in this PR:

```go
if request.Body != nil {
    closer := request.Body
    request.Body = ioutil.NopCloser(request.Body)
    defer func() {
        closer.Close()
    }()
}
```

- `(*http.Request).Body` is an `io.ReadCloser`: https://golang.org/src/net/http/request.go#L181
- By wrapping it in an `ioutil.NopCloser` the Body is never closed: https://golang.org/src/io/ioutil/ioutil.go#L114
- Which is a requirement of `RoundTripper`: https://golang.org/src/net/http/client.go#L133

rosenhouse: @judwhite glad you found this. 

but doesn't `closer.Close()` cause the original request Body to be closed?  I think that's the idea behind the variable assignment before modifying the `request.Body`

judwhite: @rosenhouse You're right, I read it wrong. I guess I'm confused why the code isn't more like:

```go
if request.Body != nil {
    defer request.Body.Close()
}
```

I suppose this prevents the passed in `io.ReadCloser` `Body` from being closed, but without looking at the bigger picture I'm not sure why.

@rosenhouse Can you also look at https://github.com/cloudfoundry/gorouter/pull/211/files#r173833484 ?

proxy/round_tripper/proxy_round_tripper.go
rosenhouse: @judwhite I agree that this *looks* overly complicated.  We just haven't figured out how to do it better.  Pull requests are welcome though....

PR #213, opened 2018-04-13
add query parameter to Opaque on //

nichtsfrei:
on path beginning with // it must set the query parameter to opaque.
Otherwise the parameter will get lost in the process.

Setting the query parameter in the URL.Opaque inside Director of Reverseproxy so that those will not getting lost in the process.

Someone hits gorouter with test.io//something?q=something than target of that should receive the complete information including the query parameter so that service can decide how to handle that.

start gorouter, start a test service, register that service and than call it.

curl 'my_first_url.localhost.routing.cf-app.com:8081/internal/status?q=hi' ends up at test service with q=hi

curl 'my_first_url.localhost.routing.cf-app.com:8081/internal/status?q=hi' ends ends up at test service without q=hi

* [ X I have made this pull request to the `master` branch

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -192,13 +192,13 @@ func (p *proxy) setupProxyRequest(target *http.Request) {
 
 	target.URL.Scheme = "http"
 	target.URL.Host = target.Host
-	target.URL.RawQuery = ""
 	target.URL.ForceQuery = false
 	target.URL.Opaque = target.RequestURI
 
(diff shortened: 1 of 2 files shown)


shalako:
Hello @nichtsfrei we noticed you closed this already. What happened?

nichtsfrei:
Oh, sty, I thought I was supposed to close and reopen it because the cla arrived after I created the pr?

shalako:
No, you did the right thing. Thank you.

PR #214, opened 2018-04-13
add query parameter to Opaque on //

nichtsfrei:
path beginning with // it must set the query parameter to opaque.
Otherwise the parameter will get lost in the process.

A short explanation of the proposed change:
Setting the query parameter in the URL.Opaque inside Director of Reverseproxy so that those will not getting lost in the process.

An explanation of the use cases your change solves
Someone hits gorouter with test.io//something?q=something than target of that should receive the complete information including the query parameter so that service can decide how to handle that.

Instructions to functionally test the behavior change using operator interfaces (BOSH manifest, logs, curl, and metrics)
start gorouter, start a test service, register that service and than call it.

curl 'my_first_url.localhost.routing.cf-app.com:8081/internal/status?q=hi' ends up at test service with q=hi

curl 'my_first_url.localhost.routing.cf-app.com:8081/internal/status?q=hi' ends ends up at test service without q=hi

### DIFF
--- a/proxy/proxy.go
+++ b/proxy/proxy.go
@@ -192,13 +192,13 @@ func (p *proxy) setupProxyRequest(target *http.Request) {
 
 	target.URL.Scheme = "http"
 	target.URL.Host = target.Host
-	target.URL.RawQuery = ""
 	target.URL.ForceQuery = false
 	target.URL.Opaque = target.RequestURI
 
 	if strings.HasPrefix(target.RequestURI, "//") {
-		target.URL.Opaque = "//" + target.Host + target.URL.Path
+		target.URL.Opaque = "//" + target.Host + target.URL.Path + target.URL.Query().Encode()
 	}
+	target.URL.RawQuery = ""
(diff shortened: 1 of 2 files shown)


nichtsfrei:
Hello @shalako is there anything I can do to speed things up, like changing something or so?

shubhaat:
Hi @nichtsfrei , 

We verified that this is a valid bug and query parameters are lost when preceded by a `//` anywhere in the URL. If you could make a PR for this, we would be happy to review and accept help. 

Regards, 
Shubha

nichtsfrei:
Hi @shubhaat ,

sorry I'm a bit confused this is the pr for the fix.

By adding ` + target.URL.Query().Encode()` to target.URL.Opaque the query parameter will not be lost anymore.

Regards,
Philipp

Nino-K:
@nichtsfrei Thanks for your contribution, your PR is merge now.

@nhsieh & Nino

PR #227, opened 2018-09-13
Prune TLS endpoints when Dial fails

jszroberto:
[#158847588]

Hi, this PR implements what was decided with [Change TLS route pruning behavior on dial error](https://github.com/cloudfoundry/gorouter/blob/master/docs/decisions/0002-change-tls-route-pruning-behavior-on-dial-error.md). Let me know what you think and what's missing (including test coverage and logging info). 

### DIFF
--- a/handlers/lookup_test.go
+++ b/handlers/lookup_test.go
@@ -78,7 +78,7 @@ var _ = Describe("Lookup", func() {
 		var pool *route.Pool
 
 		BeforeEach(func() {
-			pool = route.NewPool(&route.PoolOpts{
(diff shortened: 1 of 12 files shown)


jszroberto:
As a side comment, I wasn't able to use `With(zap.Error(err))` in the pool, tests in many packages were panicking because some of the errors don't have the field "error". If there is any trick, let me know.

jszroberto:
Another point is that gorouter retries to make the request when getting Dial errors for TLS.

zachgersh:
Closing, really appreciate this! Went a slightly different direction with the implementation after talking with some folks here.

PR #230, opened 2018-10-19
Allow inject HSTS header to response

keymon:

What?
-----

A new option strict_transport_security_header is added and can be
set to any free string to be inserted in this header, the
operator must set the right format.

The header would be injected only if configured and if the backend
does not return that header.

The implementation extends ProxyResponseWriter to allow inject
any the header, and we add a new custom handler. This would make
it easier to add similar features later.

Use case
--------

In our use case we want every application to have the right HSTS
headers configured by default, as it is a requirement in our
organisation for any service.

The tenants can always override this header.

How to test/review?
-------------------

Tests where added.

To test it, you must set the `strict_transport_security_header`
config and test it by checking that the gorouter  returns
the header for any proxied HTTP request.

A fork of the routing-release with this option has been
added in: https://github.com/keymon/routing-release/tree/add_hsts_header
and will be

* [ ] I have made this pull request to the `master` branch

* [ ] I have run all the unit tests using `bin/test`

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -139,6 +139,8 @@ type Config struct {
 	IsolationSegments        []string          `yaml:"isolation_segments,omitempty"`
 	RoutingTableShardingMode string            `yaml:"routing_table_sharding_mode,omitempty"`
 
+	StrictTransportSecurityHeader string `yaml:"strict_transport_security_header,omitempty"`
+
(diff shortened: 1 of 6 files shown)


keymon:
Sorry, opened from the wrong fork. Ignore.

Right one in #231

PR #231, opened 2018-10-19
Allow inject HSTS header to response

keymon:
What?
-----

A new option strict_transport_security_header is added and can be
set to any free string to be inserted in this header, the
operator must set the right format.

The header would be injected only if configured and if the backend
does not return that header.

The implementation extends ProxyResponseWriter to allow inject
any the header, and we add a new custom handler. This would make
it easier to add similar features later.

Use case
--------

In our organisation there is a requirement of any application to use HTTPS and set HSTS headers. We want every application to respond with these headers. The tenant can always override them if they need to.

How to review/test
----------------------

 - Code review, there are tests
 - Set the right config `strict_transport_security_header` with the right value. Check that gorouter returns the header as desired.
 - A fork of routing-release with the right parameter is in: https://github.com/alphagov/paas-routing-release/tree/add_hsts_header and https://github.com/cloudfoundry/routing-release/pull/126

-------

Header added if the `strict_transport_security_header` is configured and the app does not return such header.

------------

Nothing happens :)

-------

PR for routing release: https://github.com/cloudfoundry/routing-release/pull/126

Checklist
--------

      - The tests fail for me in master anyway. There are some problems with `go vet` in `main.go` (router variable overrides package) 

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -139,6 +139,8 @@ type Config struct {
 	IsolationSegments        []string          `yaml:"isolation_segments,omitempty"`
 	RoutingTableShardingMode string            `yaml:"routing_table_sharding_mode,omitempty"`
 
+	StrictTransportSecurityHeader string `yaml:"strict_transport_security_header,omitempty"`
+
 	CipherString                      string             `yaml:"cipher_suites,omitempty"`
 	CipherSuites                      []uint16           `yaml:"-"`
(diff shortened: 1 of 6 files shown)


KauzClay:
@keymon can we close this out and refer only to the more generic PR #232 ? Or is this still necessary?

keymon:
:+1:

PR #232, opened 2018-10-28
Allow inject custom headers into response

keymon:
What?
-----

To do so here we add the posibility to inject custom headers in any response from the
go router. The header would be injected only if the backend
does not return those headers.

We allow some new configuration as in the router as:

    http_rewrite:
      inject_response_headers:
      - name: X-A-Header
        value: the-value
      - name: X-A-Header
        value: other-value
      - name: X-Other-Header
        value: some value

Scope
------

We only implement add the headers if they are not set in the backend. Other operations  (append, override, delete) can be easily added if desired.

Implementation details
-----------------------------

This commit adds the basis to implement any rewrite logic on
the headers. We extend ProxyResponseWriter to allow register
new HeaderRewriters.

We add for the time being a InjectHeaderRewriter, that would only
add headers if they don't already exist. Similar HeaderRewriters
can be implemented for delete, append, etc.

A new Handler HTTPRewrite is inserted if there is any `http_rewrite`
configuration. This handler would initialise once HeaderRewriter
during the setup of the server, and attach it to each
ProxyResponseWriter in each request.

Use case
--------

In our organisation there is a requirement of any application to use HTTPS and set HSTS headers. We want every application to respond with these headers. The tenant can always override them if they need to.

How to review/test
----------------------

 - Code review, there are tests
 - Set the right config as aboce. Check that gorouter returns the header as desired.
 - A fork of routing-release with the right parameter is in: TBD

-------

Header added if the `strict_transport_security_header` is configured and the app does not return such header.

------------

Nothing happens :)

-------

PR for routing release: TBD

Checklist
--------

      - The tests fail for me in master anyway. There are some problems with `go vet` in `main.go` (router variable overrides package) 

### DIFF
--- a/proxy/proxy_test.go
+++ b/proxy/proxy_test.go
@@ -31,6 +31,7 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
 )
 
 var _ = Describe("Proxy", func() {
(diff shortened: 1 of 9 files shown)


proxy/proxy_test.go
```diff
@@ -612,6 +613,82 @@ var _ = Describe("Proxy", func() {
 		})
 	})
 
+	FDescribe("HTTP Rewrite", func() {
```
keymon: @zachgersh this has been merged with a focus! 

I didn't expect to be merged without agreeing the right interface, sorry for that.

Shall we do a PR asap to fix thix

keymon: Fixed in e2b2187dc9af8a233005d5edecf168900e9b1e78 :+1:

zachgersh: @keymon - yeah we caught the focus and some others (no big deal). The code as implemented worked (so we were happy to not delay a merge). A fast follow PR to fix the yaml structure that @shalako wants would be great (and will be merged quickly as well).

PR #241, opened 2018-12-28
http keep-alive connections to backends outlast the route-registration interval

rosenhouse:
In reviewing our proposed fixes to #235, I realized that idle connections to backends were only surviving until the next route registration message arrived.

this PR adds integration test coverage for HTTP persistent connections (keep-alive) to backends, and introduces a (kind of hacky) fix to enable the connections to survive beyond the route registration messages.

to test, push an app which prints logs when the state of its remote connections change, and observe how those connections persist when new route registration messages are published to gorouter.


shalako:
Oh wow, good find!

PR #244, opened 2019-01-18
Disable caching for unknown_route responses

### LINKED ISSUE #243: Route not found 404s should not be cached
## Issue

gorouter doesn't return a Cache-Control header for 404s. This can cause issues for sites that sit behind HTTP caches.

## Context

If all of the apps serving a particular app fail their healthcheck, gorouter will respond with a 404.

If there are caching proxies in front of gorouter these may decide to cache this 404 when the applications become healthy again.

This can result in the undesirable situation that an application that has a few seconds of downtime can end up with a cached 404 for hours (depending on the cache configuration).

## Steps to Reproduce

```
curl -v http://some-gorouter-route-that-does-not-exist.localhost:8080
```

## Expected result

`Cache-Control: no-cache, no-store`

## Current result

No `Cache-Control` header.

## Possible Fix

Add `Cache-Control: no-cache, no-store` to route-not-found 404 responses. Leave headers for application 404s as they are.

richardTowers:
## A short explanation of the proposed change:

See #243 - it's possible that caches in front of gorouter will cache temporary "unknown route" 404s. Adding a cache-control header should prevent this.

## An explanation of the use cases your change solves

If all of the apps serving a particular app fail their healthcheck,
gorouter will respond with a 404.

If there are caching proxies in front of gorouter these may decide to
cache this 404 beyond the point when applications become healthy again.

This can result in the undesirable situation that an application that
has a few seconds of downtime can end up with a cached 404 for hours
(depending on the cache configuration).

By setting `Cache-Control: no-cache` on unknown_route responses we can
instruct any caches in front of gorouter to always check that the
response is current before deciding to serve a 404 from their cache.

We could consider setting `no-store` instead, which would prevent all
caching, but `no-cache` should be enough to prevent any issues.

## Instructions to functionally test the behavior change using operator interfaces (BOSH manifest, logs, curl, and metrics):

`curl http://asldkfjaldkjfalsdk.some-gorouter`

## Expected result after the change

Unknown route responses should contain the `Cache-Control: no-cache` header

## Current result before the change

Unknown route responses contain no caching headers

## Links to any other associated PRs

Fixes #243

### DIFF
--- a/handlers/lookup.go
+++ b/handlers/lookup.go
@@ -60,6 +60,7 @@ func (l *lookupHandler) handleMissingRoute(rw http.ResponseWriter, r *http.Reque
 	l.reporter.CaptureBadRequest()
 
 	rw.Header().Set("X-Cf-RouterError", "unknown_route")
+	rw.Header().Set("Cache-Control", "public,max-age=2")
 
 	writeStatus(
 		rw,
--- a/handlers/lookup_test.go
+++ b/handlers/lookup_test.go
@@ -64,6 +64,10 @@ var _ = Describe("Lookup", func() {
 			Expect(resp.Header().Get("X-Cf-RouterError")).To(Equal("unknown_route"))
 		})
 
+		It("Sets Cache-Control to public,max-age=2", func() {
+			Expect(resp.Header().Get("Cache-Control")).To(Equal("public,max-age=2"))
+		})
+
 		It("returns a 404 NotFound and does not call next", func() {
 			Expect(nextCalled).To(BeFalse())
 			Expect(resp.Code).To(Equal(http.StatusNotFound))


handlers/lookup.go
```diff
@@ -60,6 +60,7 @@ func (l *lookupHandler) handleMissingRoute(rw http.ResponseWriter, r *http.Reque
 	l.reporter.CaptureBadRequest()
 
 	rw.Header().Set("X-Cf-RouterError", "unknown_route")
+	rw.Header().Set("Cache-Control", "no-cache")
```
jpluscplusm: I wonder if a wider set of "do not cache" values in the `Cache-Control` header might be suitable for this default 404 scenario: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Cache_response_directives

richardTowers: I think using `no-store` instead might be a good bet.

Brief summary:
* `no-store` tells caches not to cache this resource at all.
* `no-cache` tells caches to always check that the E-Tag with upstream, but they're allowed to respond from cache if they get a 304 Not-Modified back.

I can't imagine gorouter is going to be serving 304s for unknown-routes, and in any case the payload is tiny in this case, so probably just disabling caching entirely is the right thing to do.

I think there's no point in adding `must-revalidate` and friends, because they're already covered by `no-cache` / `no-store`.

richardTowers: Updated to use `no-store`.

PR #220, opened 2018-07-09
Do not set VCAP_ID stickyness cookie if value already provided by backend. 

thomasbuengener:
…kend.

Thanks for contributing to gorouter. To speed up the process of reviewing your pull request please provide us with:

* A short explanation of the proposed change:
Do not set VCAP_ID stickyness cookie if value already provided by backend. 
GoRoute would set the VCAP_ID Cookie ignoring an already present Set-Cookie Header from the backend, which results in a dublicate Set-Cookie Header.
This is important to allow the backend to change the routing of a sticky session to a different instance.

* An explanation of the use cases your change solves
Rerouting indication from the backend where overwritten by gorouter. gorouter schould not set the 
VCAP_ID Cookie twice if it is already present in the backend response.

* Instructions to functionally test the behavior change using operator interfaces (BOSH manifest, logs, curl, and metrics)

* Expected result after the change

* Current result before the change

* Links to any other associated PRs

* [X] I have viewed signed and have submitted the Contributor License Agreement

* [X] I have made this pull request to the `master` branch

* [ ] I have run all the unit tests using `bin/test`

* [ ] I have run CF Acceptance Tests on bosh lite

PR #245, opened 2019-02-11
Do not set VCAP_ID stickyness cookie if value already provided by backend. 

thomasbuengener:
See
https://github.com/cloudfoundry/gorouter/pull/220

I could no reopen the Request #220 so I had to created this new one.
I updated the missing test from #220 now.

Do not set VCAP_ID stickyness cookie if value already provided by backend. 
GoRoute would set the VCAP_ID Cookie ignoring an already present Set-Cookie Header from the backend, which results in a dublicate Set-Cookie Header.
This is important to allow the backend to change the routing of a sticky session to a different instance.

Rerouting indication from the backend where overwritten by gorouter. gorouter schould not set the 
VCAP_ID Cookie twice if it is already present in the backend response.

* [ ] I have run all the unit tests using `bin/test`

### DIFF
--- a/proxy/round_tripper/proxy_round_tripper.go
+++ b/proxy/round_tripper/proxy_round_tripper.go
@@ -310,6 +310,13 @@ func setupStickySession(
 			break
 		}
 	}
+	
+	for _, v := range response.Cookies() {
+		if v.Name == VcapCookieId {
+			sticky = false
+			break
+		}
+	}	
 
 	if sticky {
 		// right now secure attribute would as equal to the JSESSION ID cookie (if present),
--- a/proxy/round_tripper/proxy_round_tripper_test.go
+++ b/proxy/round_tripper/proxy_round_tripper_test.go
@@ -874,6 +874,69 @@ var _ = Describe("ProxyRoundTripper", func() {
 				})
 			})
 		})
+		
+
+		Context("when sticky session vcap cookie provided by backend", func() {
+				var (
+						sessionCookie *http.Cookie
+						vcapCookie    *http.Cookie
+						endpoint1     *route.Endpoint
+						endpoint2     *route.Endpoint
+				)
+
+				BeforeEach(func() {
+						sessionCookie = &http.Cookie{
+								Name: round_tripper.StickyCookieKey,
+						}
+
+						vcapCookie = &http.Cookie{
+								Name: round_tripper.VcapCookieId,
+						}
+
+						transport.RoundTripStub = func(req *http.Request) (*http.Response, error) {
+								resp := &http.Response{StatusCode: http.StatusTeapot, Header: make(map[string][]string)}
+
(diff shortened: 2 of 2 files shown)


shubhaat:
@thomasbuengener - would you mind signing the CLA?

chipchilders:
@shubhaat We have received his CLA. Feel free to merge!

tombueng:
I submitted the CLA. But I#m uanble to close or reopen the pull request (Simply beacuse there is no button here for that).

ameowlia:
Hi @thomasbuengener and @tombueng, 

We're looking at this with some fresh eyes and we have a question: 

It sounds like you want instanceA to be able to tell a client to talk to instanceB next time. Could you use different apps to achieve the same result? 

For example: appA and appB could each have their own route. Also, there could be one route mapped to both apps. This way, if appA wants the client to talk to appB next, traffic could be forwarded to the appB-route. This would allow you to use CF primitives, rather than messing around with the VCAP_ID.

Does this sounds like a solution you could use? 

Thanks, 
@ameowlia and @bruce-ricard, CF Networking Program Team members

tombueng:
Dear Amelia,

 

I’m not really getting how this could work out.

Could you please explain in more technical detail what

appA should do to redirect traffic to appB on the next request?

Which CF primitives should we use there?

 

Yours,

Thomas

 

 

Von: Amelia Downs <notifications@github.com> 
Gesendet: Montag, 6. Mai 2019 23:18
An: cloudfoundry/gorouter <gorouter@noreply.github.com>
Cc: Thomas Büngener <thomas@buengener.de>; Mention <mention@noreply.github.com>
Betreff: Re: [cloudfoundry/gorouter] Do not set VCAP_ID stickyness cookie if value already provided by backend. (#245)

 

Hi @thomasbuengener <https://github.com/thomasbuengener>  and @tombueng <https://github.com/tombueng> ,

We're looking at this with some fresh eyes and we have a question:

It sounds like you want instanceA to be able to tell a client to talk to instanceB next time. Could you use different apps to achieve the same result?

For example: appA and appB could each have their own route. Also, there could be one route mapped to both apps. This way, if appA wants the client to talk to appB next, traffic could be forwarded to the appB-route. This would allow you to use CF primitives, rather than messing around with the VCAP_ID.

Does this sounds like a solution you could use?

Thanks,
@ameowlia <https://github.com/ameowlia>  and @bruce-ricard <https://github.com/bruce-ricard> , CF Networking Program Team members

KauzClay:
Hey @thomasbuengener , we talked some more about this with the team. @ameowlia's solution basically involves making separate apps, using their routes in the app response to orchestration the redirection you want. But after more discussion, we think your change will work. We ran it through some acceptance tests locally and things seemed fine, so we are gonna merge this in. Sorry for such a long delay, and thanks for your contribution!

PR #247, opened 2019-02-15
fix url parsing for forwarded url after route-service pass

### LINKED ISSUE #246: Failed to validate Route Service Signature
## Issue

Gorouter throws error `Failed to validate Route Service Signature` when performing requests containing escaped chars #%.

## Context

We are currently using a route-service (with the [User-Provided Service mode](https://docs.cloudfoundry.org/services/route-services.html#user-provided)). The second time the request reaches the gorouter, it throws the error `Failed to validate Route Service Signature`.

## Steps to Reproduce

1. Trigger a request that contains escaped chars #%. E.g.: https://myapp.mydomain/?filter=somequery%23%25
2. Request reaches gorouter, which forwards request to route-service
3. Route-service perform some operations (no changes in headers), then reads the URL from `X-CF-Forwarded-URL` and forwards the request to that URL
4. Request reaches go-router again and issue happens

## Expected result

Request should go through without issues, as it's a valid request

## Current result

Error `Failed to validate Route Service Signature`

## Possible Fix

In the `validateRouteServicePool` function, that URL is retrieved and parsed, as seen [here](https://github.com/cloudfoundry/gorouter/blob/master/handlers/routeservice.go#L183). As the URL is not valid (contains a #), the `url.Parse` function throws an error.

By using the function `url.ParseRequestURI`, it seems we were able to fix it. We are doing some more tests and then will open a PR.

otaviojacobi:
### A short explanation of the proposed change:
See #246 - This PR changes the [url.Parse](https://golang.org/pkg/net/url/#Parse) to [url.ParseRequestURI](https://golang.org/pkg/net/url/#ParseRequestURI) in the [handlers.validateRouteServicePool](https://github.com/cloudfoundry/gorouter/blob/master/handlers/routeservice.go#L183) method. The main reason for that is that url.Parse tries to strip out the # from the unescaped ForwardedUrl which when followed by a % character  results in an error. The url.ParseRequestURI doesn't have this problem since it considers that the request doesn't have a #fragment.

### An explanation of the use cases your change solves
Requests that pass through the gorouter and have a bound route-service, during the second pass through the gorouter, if the request has a escaped character # followed by a escaped % (as `https://my-cf-route.my.cf.domain/test?findById='%23%25'`) throw an `Invalid Signature Error` but the request is fine and should be properly forwarded to the backing service.

### Instructions to functionally test the behavior change using operator interfaces (BOSH manifest, logs, curl, and metrics)
- create a simple backend service to be routed.
- create a route-service that simply reverse-proxy the request to the backend service
- create a user provided service from the created route-service
- bind the user provided service to the simple backend service
- make a request with a query as `?testquery=%23%25`  

### Expected result after the change
All requests with well formed URL should pass through the gorouter with a route-service and the request above should be forwarded to the backend service 

### Current result before the change
Impossible to make requests that have a escaped character `#`(%23) followed by a escaped character `%`(%25) as the gorouter throws an `Invalid Signature Error`

### Links to any other associated PRs
This fixes #246 

* [ ] I have run CF Acceptance Tests on _bosh lite_

### DIFF
--- a/handlers/routeservice_test.go
+++ b/handlers/routeservice_test.go
@@ -306,6 +306,32 @@ var _ = Describe("Route Service Handler", func() {
 				})
 			})
 
+			Context("when a request has a valid route service signature and metadata header and URL contains special chars", func() {
+				BeforeEach(func() {
+					reqArgs, err := config.Request("", "https://my_host.com/resource+9-9_9?query=%23%25")
+					Expect(err).ToNot(HaveOccurred())
+					req.Header.Set(routeservice.HeaderKeySignature, reqArgs.Signature)
+					req.Header.Set(routeservice.HeaderKeyMetadata, reqArgs.Metadata)
+				})
+
+				It("strips headers and sends the request to the backend instance", func() {
+					handler.ServeHTTP(resp, req)
+
+					Expect(resp.Code).To(Equal(http.StatusTeapot))
+
+					var passedReq *http.Request
+					Eventually(reqChan).Should(Receive(&passedReq))
+
+					Expect(passedReq.Header.Get(routeservice.HeaderKeySignature)).To(BeEmpty())
+					Expect(passedReq.Header.Get(routeservice.HeaderKeyMetadata)).To(BeEmpty())
+					Expect(passedReq.Header.Get(routeservice.HeaderKeyForwardedURL)).To(BeEmpty())
+					reqInfo, err := handlers.ContextRequestInfo(passedReq)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(reqInfo.RouteServiceURL).To(BeNil())
+					Expect(nextCalled).To(BeTrue(), "Expected the next handler to be called.")
+				})
+			})
+
 			Context("when a request has a route service signature but no metadata header", func() {
 				BeforeEach(func() {
 					reqArgs, err := config.Request("", forwardedUrl)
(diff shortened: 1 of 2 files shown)


handlers/routeservice_test.go
```diff
@@ -306,6 +306,32 @@ var _ = Describe("Route Service Handler", func() {
...
+					reqArgs, err := config.Request("", "https://my_host.com/resource+9-9_9?query=%23%25")
+					Expect(err).ToNot(HaveOccurred())
+					req.Header.Set(routeservice.HeaderKeySignature, reqArgs.Signature)
+					req.Header.Set(routeservice.HeaderKeyMetadata, reqArgs.Metadata)
+				})
+
+				It("strips headers and sends the request to the backend instance", func() {
+					handler.ServeHTTP(resp, req)
+
+					Expect(resp.Code).To(Equal(http.StatusTeapot))
```
bruce-ricard: It seems like everything passed this code test could be removed. This is really what we are testing. All the rest if just the same as the test just above. We don't need to test again that it "strips headers and sends the request to the backend instance", but just that there are no errors.

otaviojacobi: My idea for the other checks was that we assure that we were not breaking anything from the test above when using the special sequence of characters. But I can see why it is test duplication, therefore I'll remove the other checks.

otaviojacobi: Hi, I have updated the PR so now the test doesn't test again the same as the above, but only the necessary.

Issue #171, opened 2017-05-04
Customizable error pages (instead of 404 route dous not exist)


jtwaleson:
## Issue

There is no user friendly page when an app is stopped.

## Context

We're building a platform based on Cloud Foundry, and would like to show our users something else than the default "Requested route (xyz) does not exist.".

Note: we have complete control over the clusters, users do not work with cf cli themselves.

## Expected result

Per case:
* app does not exist -> generic error page
* app exists but is not running -> user definable error page

## Current result

"Requested route (xyz) does not exist."

## Possible Fix

I think this is completely beyond the scope of the gorouter, but I think these two pragmatic solutions would allow us to build this functionality with a custom app.

### 1 - wildcard routes also match the main domain

This could be solved with a custom app that returns error pages with a wildcard route. This works, because if the app is stopped, the app with the wildcard route will get the traffic, that can serve the standard or custom error pages. Unfortunately, this fails for custom domains such as `mydomain.com` without a hostname. In that case, the wildcard `*.mydomain.com` does not match if the app is stopped.

In this case, we simply add a wildcard route for each custom domain to the "custom-error-pages" app.

### 2 - a global fallback route
The possibility to register a cluster-wide fallback route, that app will get all the non-matching traffic, and can decide if it returns 404, 503, etc.

aaronshurley:
@jtwaleson We do not currently support customizable 404 pages. I think this feature may surpass the responsibilities of the gorouter but know that others have asked about this before. @shalako do we have a recommended approach for this?

-Aaron && @belinda-liu

shalako:
@jtwaleson and I spoke about this request at CF Summit. While the feature seems quite reasonable, I don't have enough evidence of market value to prioritize it, when compared to features that are in greater demand.

It doesn't appear that we could apply the requested behavior consistently across the tree of domains, as this would not be backwards compatible. Although the use case appears valid, I'm concerned with a exception to behavior for root domains as this sacrifices predictability.

Currently a request for `foo.bar.com` will match wildcard route `*.bar.com` if there is no route `foo.bar.com`, or the app for it is down. 

The requested feature would route requests for `foo.bar.com` to the route `*.foo.bar.com` if there is no route `foo.bar.com` or the app is down. But what if there is also a wildcard route for `*.bar.com`? Requests for `foo.bar.com` are currently routed to `*.bar.com` and with the proposed change we would send them to `*.foo.bar.com`. That would be a backward incompatible change.

[...]

Issue #135, opened 2016-06-17
404 in case of route not found


MichaelStephan:
Hello,

in case I want to access an app which is not registered or which is not running at a certain time the router returns a 404. Unfortunately 404 may also have a business meaning next to the technical issue indicated by the router. Could the cloud foundry router either return 503 (which is a breaking change and therefor I would understand that it cannot simply be done) or indicate with a custom header field that the 404 was returned by the router itself (router error) and no upstream service.

Best regards,
Michael
(michael.stephan@sap.com)

crhino:
Hi @MichaelStephan,

There has been a number of discussion on the mailing list about this topic. In summary, we might consider returning a 503 status code for routes that are registered on Cloud Foundry but do not have any applications currently mapped to them, but that would involve some substantial changes to how routes are registered.

There are a few workarounds that people have come with, see the links below for more details on that and discussion around this issue.

[CF RTR returning 404 when app is down](https://lists.cloudfoundry.org/archives/list/cf-dev@lists.cloudfoundry.org/thread/CO6S5HLHQCJRS4ULZUUCFBE5YJRHDAWZ/#DRWZNGR6Q43LSWMORQJ66J25JHGGUYWA)
[How can we customized "404 Not Found"](http://cf-dev.70369.x6.nabble.com/cf-dev-How-can-we-customized-quot-404-Not-Found-quot-td4501.html#a4913)
[Brokered route services only receiving traffic for routes mapped to started apps](http://cf-dev.70369.x6.nabble.com/cf-dev-Brokered-route-services-only-receiving-traffic-for-routes-mapped-to-started-apps-tp4699p4742.html)

shashwathi:
@MichaelStephan : In the description you mentioned custom header field to truly represent `404` returned by the router itself. There is a header field`X-Cf-RouterError` which is set to `unknown_route` by gorouter.

Regards
Shash

swetharepakula:
@MichaelStephan. We are closing this for now due to inactivity. Please let us know if you would like to reopen for further discussion.

Thanks,
@swetharepakula && @aaronshurley

Netha6571:
@MichaelStephan did you get any solution for this problem.

PR #251, opened 2019-05-31
Return a 503 not a 404 when no endpoint available

tlwr:
Context
---

Currently gorouter returns a 404 when a route is known but there is nothing in the pool available to handle the request.

HTTP 4xx is for user related problems, HTTP 5xx is for server related problems.

An empty pool is incorrect platform behaviour. HTTP `503 ServiceUnavailable` which accurately describes this behaviour.

Change
---

Instead of returning a `404 NotFound` when the pool `IsEmpty`, return a `503 ServiceUnavailable`.

Why
---

We currently have to differentiate whether the 404 is:
> A real lack of a route

or

> Nothing being available currently to handle the route

by looking in gorouter logs, checking whether there is an upstream (e.g. `"10.x.x.x:xxxxx"` vs `"-"`)

It would be better if this was a `503` because it more accurately describes the situation.

Decisions to make
---

- Record a metric, what metric should be recorded, do we need a new one?
- Does this behaviour need to be configurable?
- Is `503` the correct status code? `521 WebServerIsDown` is a Cloudflare status code (like `526` which is also used by gorouter) which more accurately describes this situation.

---

***THIS IS A DRAFT***

### DIFF
--- a/handlers/helpers.go
+++ b/handlers/helpers.go
@@ -5,9 +5,25 @@ import (
 	"net/http"
 	"strings"
 
+	router_http "code.cloudfoundry.org/gorouter/common/http"
 	"code.cloudfoundry.org/gorouter/logger"
 	"github.com/uber-go/zap"
 )
+
+const (
+	cacheMaxAgeSeconds = 2
+)
+
+func AddRouterErrorHeader(rw http.ResponseWriter, val string) {
+	rw.Header().Set(router_http.CfRouterError, val)
+}
+
+func addInvalidResponseCacheControlHeader(rw http.ResponseWriter) {
+	rw.Header().Set(
+		"Cache-Control",
+		fmt.Sprintf("public,max-age=%d", cacheMaxAgeSeconds),
+	)
+}
 
 func writeStatus(rw http.ResponseWriter, code int, message string, logger logger.Logger) {
 	body := fmt.Sprintf("%d %s: %s", code, http.StatusText(code), message)
--- a/handlers/lookup.go
+++ b/handlers/lookup.go
@@ -18,8 +18,6 @@ import (
 const (
 	CfInstanceIdHeader = "X-CF-InstanceID"
 	CfAppInstance      = "X-CF-APP-INSTANCE"
-
-	cacheMaxAgeSeconds = 2
 )
 
 type lookupHandler struct {
@@ -61,11 +59,8 @@ func (l *lookupHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request, next 
 func (l *lookupHandler) handleMissingRoute(rw http.ResponseWriter, r *http.Request) {
 	l.reporter.CaptureBadRequest()
 
-	rw.Header().Set("X-Cf-RouterError", "unknown_route")
-	rw.Header().Set(
-		"Cache-Control",
-		fmt.Sprintf("public,max-age=%d", cacheMaxAgeSeconds),
-	)
+	AddRouterErrorHeader(rw, "unknown_route")
+	addInvalidResponseCacheControlHeader(rw)
 
 	writeStatus(
 		rw,
@@ -79,7 +74,7 @@ func (l *lookupHandler) handleOverloadedRoute(rw http.ResponseWriter, r *http.Re
 	l.reporter.CaptureBackendExhaustedConns()
 	l.logger.Info("connection-limit-reached")
 
-	rw.Header().Set("X-Cf-RouterError", "Connection Limit Reached")
+	AddRouterErrorHeader(rw, "Connection Limit Reached")
 
 	writeStatus(
 		rw,
--- a/handlers/lookup.go
+++ b/handlers/lookup.go
@@ -37,8 +37,13 @@ func NewLookup(registry registry.Registry, rep metrics.ProxyReporter, logger log
 
 func (l *lookupHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
 	pool := l.lookup(r)
-	if pool == nil || pool.IsEmpty() {
+	if pool == nil {
 		l.handleMissingRoute(rw, r)
+		return
+	}
+
+	if pool.IsEmpty() {
+		l.handleUnavailableRoute(rw, r)
 		return
 	}
 
@@ -66,6 +71,18 @@ func (l *lookupHandler) handleMissingRoute(rw http.ResponseWriter, r *http.Reque
 		rw,
 		http.StatusNotFound,
 		fmt.Sprintf("Requested route ('%s') does not exist.", r.Host),
+		l.logger,
+	)
+}
+
+func (l *lookupHandler) handleUnavailableRoute(rw http.ResponseWriter, r *http.Request) {
+	AddRouterErrorHeader(rw, "no_endpoints")
+	addInvalidResponseCacheControlHeader(rw)
+
+	writeStatus(
+		rw,
+		http.StatusServiceUnavailable,
+		fmt.Sprintf("Requested route ('%s') has no available endpoints.", r.Host),
 		l.logger,
 	)
 }
--- a/handlers/lookup_test.go
+++ b/handlers/lookup_test.go
@@ -60,11 +60,11 @@ var _ = Describe("Lookup", func() {
 			Expect(rep.CaptureBadRequestCallCount()).To(Equal(1))
 		})
 
-		It("Sets X-Cf-RouterError to unknown_route", func() {
+		It("sets X-Cf-RouterError to unknown_route", func() {
 			Expect(resp.Header().Get("X-Cf-RouterError")).To(Equal("unknown_route"))
 		})
 
-		It("Sets Cache-Control to public,max-age=2", func() {
+		It("sets Cache-Control to public,max-age=2", func() {
 			Expect(resp.Header().Get("Cache-Control")).To(Equal("public,max-age=2"))
 		})
 
@@ -96,7 +96,7 @@ var _ = Describe("Lookup", func() {
 			Expect(rep.CaptureBadRequestCallCount()).To(Equal(1))
 		})
 
-		It("Sets X-Cf-RouterError to unknown_route", func() {
+		It("sets X-Cf-RouterError to unknown_route", func() {
 			Expect(resp.Header().Get("X-Cf-RouterError")).To(Equal("unknown_route"))
 		})
 
--- a/handlers/lookup_test.go
+++ b/handlers/lookup_test.go
@@ -92,23 +92,26 @@ var _ = Describe("Lookup", func() {
 			reg.LookupReturns(pool)
 		})
 
-		It("sends a bad request metric", func() {
-			Expect(rep.CaptureBadRequestCallCount()).To(Equal(1))
+		It("does not send a bad request metric", func() {
+			Expect(rep.CaptureBadRequestCallCount()).To(Equal(0))
 		})
 
-		It("sets X-Cf-RouterError to unknown_route", func() {
-			Expect(resp.Header().Get("X-Cf-RouterError")).To(Equal("unknown_route"))
+		It("sets X-Cf-RouterError to no_endpoints", func() {
+			Expect(resp.Header().Get("X-Cf-RouterError")).To(Equal("no_endpoints"))
 		})
 
-		It("returns a 404 NotFound and does not call next", func() {
+		It("returns a 503 ServiceUnavailable and does not call next", func() {
 			Expect(nextCalled).To(BeFalse())
-			Expect(resp.Code).To(Equal(http.StatusNotFound))
+			Expect(resp.Code).To(Equal(http.StatusServiceUnavailable))
 		})
 
 		It("has a meaningful response", func() {
-			Expect(resp.Body.String()).To(ContainSubstring("Requested route ('example.com') does not exist"))
+			Expect(resp.Body.String()).To(ContainSubstring("Requested route ('example.com') has no available endpoints"))
 		})
 
+		It("Sets Cache-Control to public,max-age=2", func() {
+			Expect(resp.Header().Get("Cache-Control")).To(Equal("public,max-age=2"))
+		})
 	})
 
 	Context("when there is a pool that matches the request, and it has endpoints", func() {
--- a/accesslog/schema/access_log_record_test.go
+++ b/accesslog/schema/access_log_record_test.go
@@ -57,7 +57,7 @@ var _ = Describe("AccessLogRecord", func() {
 	})
 
 	Describe("LogMessage", func() {
-		It("Makes a record with all values", func() {
+		It("makes a record with all values", func() {
 			r := BufferReader(bytes.NewBufferString(record.LogMessage()))
 			Eventually(r).Should(Say(`FakeRequestHost\s-\s\[\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}\+\d{4}\]`))
 			Eventually(r).Should(Say(`"FakeRequestMethod http://example.com/request FakeRequestProto" `))
@@ -100,7 +100,7 @@ var _ = Describe("AccessLogRecord", func() {
 				}
 			})
 
-			It("Makes a record with all values", func() {
+			It("makes a record with all values", func() {
 				r := BufferReader(bytes.NewBufferString(record.LogMessage()))
 				Eventually(r).Should(Say(`FakeRequestHost\s-\s\[\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}\+\d{4}\]`))
 				Eventually(r).Should(Say(`"FakeRequestMethod http://example.com/request FakeRequestProto" `))
@@ -169,7 +169,7 @@ var _ = Describe("AccessLogRecord", func() {
 		})
 
 		Context("when extra headers is an empty slice", func() {
-			It("Makes a record with all values", func() {
+			It("makes a record with all values", func() {
 				record := schema.AccessLogRecord{
 					Request: &http.Request{
 						Host:   "FakeRequestHost",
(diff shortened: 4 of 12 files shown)


tlwr:
My colleague @richardTowers pointed me at https://github.com/cloudfoundry/gorouter/issues/135, which is relevant but the codebase has moved on a bit.

AFAIK Istio using `istio-release` returns a `503` when there is nothing to service a request but the route exists, so this behaviour would be consistent.

richardTowers:
Not a contributor, but for what it's worth here are my thought's on @tlwr's questions (full disclosure: we're colleagues):

> Record a metric, what metric should be recorded, do we need a new one?

Yes, I think a new metric would be good. Guessing something similar to `l.reporter.CaptureBadRequest()`

> Does this behaviour need to be configurable?

Not for PaaS. It would be less of a breaking change if it was configurable and defaulted to 404, but adding configuration has a cost in terms of complexity that probably isn’t worth paying for this.

> Is 503 the correct status code? 521 WebServerIsDown is a Cloudflare status code (like 526 which is also used by gorouter) which more accurately describes this situation.

I think 503 is better, although I can see that since they already use some non-standard codes it would be consistent to use another one. I think 521 might not be the right choice though, because CloudFlare has a specific and slightly different meaning (“the backend refused the connection”, rather than “there weren’t any healthy backends to serve your request”)

tlwr:
I think these are reasonable answers to the questions I raised, I would like someone from the cf-networking team to weigh in. In the meantime I'll have a crack at implementing the above decisions

tlwr:
I've thought about this some more, I think Gorouter shouldn't use `4xx` for either case where it currently does, although this is a sufficiently controversial / breaking change that it would be better for it to be configurable, which I will attempt.

The behaviour I want from gorouter (this won't apply to everyone) is:

> When the pool is empty, return a `521 Web Server Is Down`

and

> When a route is not found, return a `523 Origin Is Unreachable`

cameronmoreau:
Hey @tlwr and @richardTowers 
Thanks for opening the PR and asking these questions. We've prioritized this issue and are currently doing a bit of research on our end. We plan on getting back to you very soon!

cc @barakyo

barakyo:
Hi @tlwr,

Just out of curiosity, is there a specific case or steps to reproduce to get into this situation that you're trying to address?

I'm asking because we were discussing when this situation would occur. We'd like to better understand that situation so we can determine how the GoRouter should address this error.

Thanks!

tlwr:
Hi @barakyo 

Right now Gorouter and upstream applications share the same HTTP response codes. Therefore an application serving a 502 or a 404 is indistinguishable from Gorouter serving a 404.

As a Cloud Foundry operator I would like to be able to distinguish platform level failure from an application level failure easily, and I think a good way of doing this would be using "platform" HTTP codes, like Cloudflare does.

An application serving a 404 is a valid HTTP response code, however this is hard to distinguish from a "route not found" in Gorouter, which is also hard to distinguish from "route exists but there are no apps to service the request". In order to distinguish the latter two cases we currently parse gorouter logs.

I think this situation could be made better by Gorouter using the 4xx response space at all, as gorouter cannot know before forwarding the request if the request is acceptable or not. Instead of returning a 404, Gorouter should return a 521 or 523 depending on whether there are application instances in the pool or not.

Therefore, as an operator looking at Gorouter metrics:
- if I see a 404, it is most likely normal behaviour of a tenant application
- if I see a 502, it is most likely a tenant reverse proxy timing out (e.g. auth proxy)
- if I see a 523, it is a tenant application taking too long to service the request
- if I see a 521, a tenant application has registered a route, but there is nothing in the pool to service the request

Although, as above, I don't expect this behaviour to be desirable for everyone, but it would help in our use case

KauzClay:
Hey @tlwr, thanks for the detailed answer. I think we are on the same page as far as the meaning of the error codes now. Before we decide on the appropriate status, we would like to know an example of where this error code would be valuable for the operator. Could you describe a case where the pool is non-nil and empty and what that means from the operator/user point of view? For context, a pool is removed from the [registry when the last endpoint is pruned](https://github.com/cloudfoundry/gorouter/blob/38191a03246cbf9ded5cb2b51c94629b2e1de94d/registry/registry.go#L159).

tlwr:
Hi, sorry for the slow response.

---

HTTP 521

AFAICT the pool will be present but empty when there are no healthy instances, for instance when an application takes too long to respond to HTTP healthchecks.

This is valuable information when assisting tenants in debugging their applications:
> why did I get 5xxs during my load test but never receive the request in my application logs

For this particular case we end up spending a lot of time in Kibana looking for Gorouter 502s with no upstream address in the logs. Ideally this would be a metric instead, which is separate from regular 5xxs.

---

HTTP 523

Some of our tenants make extensive use of internal applications. There will be a reverse proxy with an external route, e.g. nginx or an OAuth proxy. When applications behind the proxies take too long to respond, the proxies themselves will return a 502.

We spend some helping our tenants investigate timeouts which is often just their applications behind a proxy. As an operator it would be valuable to distinguish platform level timeouts from tenants running their own proxies.

Tenants could use a route service instead, but for various reasons do not want to.

---

Lastly, having Gorouter return 521s and 523s would help operators implement their own custom error pages without Gorouter having this functionality. An operator could run a proxy in front of Gorouter which inspected the returned HTTP status code, if it saw a platform level error, it would return a custom HTML page which could be more user friendly than the current `404 Not Found: Requested route ('foobar.cloudapps.digital') does not exist.`, which would be steps towards #171.

(This last point is irrelevant though so please feel free to ignore)

KauzClay:
Hi @tlwr , thanks for the explanation. 

We agree there could be more information provided upon error, but we don't feel that the more custom error codes is the way to go. Instead, how about using a 503 and then adding more informative headers at different errors? Could inspecting the response headers give you the same benefit? Other errors use this pattern (Ex: when an endpoint is unavailable, you get a 503 with the header `X-Cf-Routererror: endpointd_failure`)

tlwr:
If "changing the error codes is out of the question" is an axiom then HTTP 503 instead of 404 makes sense and I endorse this change.

I dislike the pattern for overloading a status code that occurs in two distinct situations, and differentiating using a header, because that does not help retroactively when looking in aggregate (unless you log all the X-CF-* headers). IMO distinct status codes is a better design but will probably cause problems for other operators.

KauzClay:
Thanks for going along with the suggestion @tlwr, those make sense to me.

mike1808:
Hi @tlwr 

Is this something you are currently working on, or we can take over and finalize it? 

Thanks,
@mike1808 and @rodolfo2488, CF Networking Program Members.

tlwr:
Hi, thanks for the nudge, and sorry for taking so long; marking it as ready for review.

As of this PR:

- (no change) If there is a timeout - return a 504 with the `X-Cf-Routererror: endpoint_failure`
- (change) If there is nothing in the pool to handle the request:
  - return a 503 immediately, without calling next
  - set header: `X-Cf-Routererror: no_endpoints`
  - set header: `Cache-Control: public,max-age=2`

I've added an assertion that it _does not_ report a `BadRequest`, but _have not_ added a metric that it does report on. I think it's probably best if you folks take this on and finalise it.

Although I'm happy to help out if there is more you need from me, thanks again

handlers/lookup.go
```diff
@@ -75,6 +80,21 @@ func (l *lookupHandler) handleMissingRoute(rw http.ResponseWriter, r *http.Reque
 	)
 }
 
+func (l *lookupHandler) handleUnavailableRoute(rw http.ResponseWriter, r *http.Request) {
+	rw.Header().Set("X-Cf-RouterError", "no_endpoints")
```
mike1808: The code here looks similar to the code from `handleMissingRoute`. Maybe we can extract that logic?

handlers/lookup_test.go
```diff
@@ -92,23 +92,26 @@ var _ = Describe("Lookup", func() {
...
 
-		It("sends a bad request metric", func() {
-			Expect(rep.CaptureBadRequestCallCount()).To(Equal(1))
+		It("does not send a bad request metric", func() {
+			Expect(rep.CaptureBadRequestCallCount()).To(Equal(0))
 		})
 
-		It("Sets X-Cf-RouterError to unknown_route", func() {
-			Expect(resp.Header().Get("X-Cf-RouterError")).To(Equal("unknown_route"))
+		It("Sets X-Cf-RouterError to no_endpoints", func() {
```
mike1808: I know that this was in the original code, but can you please change `Sets` to `sets`?

handlers/lookup_test.go
```diff
@@ -92,23 +92,26 @@ var _ = Describe("Lookup", func() {
...
+			Expect(resp.Header().Get("X-Cf-RouterError")).To(Equal("no_endpoints"))
 		})
 
-		It("returns a 404 NotFound and does not call next", func() {
+		It("returns a 503 NotFound and does not call next", func() {
```
mike1808: You change the status code but not the message associated with the code.

mike1808:
@tlwr thank you for finalizing! 

Regarding `BadRequest` metric, @keshav-pivotal @adobley do you think we can safely not emit that metric when there no endpoints registered for a route?

tlwr:
Cool I'll action these comments

handlers/lookup_test.go
tlwr: Resolved in https://github.com/cloudfoundry/gorouter/pull/251/commits/8c0eb34791893c96976965c37237241d639c4c49

handlers/lookup_test.go
tlwr: Done in https://github.com/cloudfoundry/gorouter/pull/251/commits/8c0eb34791893c96976965c37237241d639c4c49

handlers/lookup.go
tlwr: Done in https://github.com/cloudfoundry/gorouter/pull/251/commits/5936addabfcf0061c4c5ebe61ddf1239389b0d19

tlwr:
Addressed the comments above

I refactored out some more of the header setting in https://github.com/cloudfoundry/gorouter/pull/251/commits/5936addabfcf0061c4c5ebe61ddf1239389b0d19 but some of the tests are still checking for the header string `"X-Cf-RouterError"` instead of using the exported const strong `CfRouterError` from code.cloudfoundry.org/gorouter/common/http ; however it is DRYer than before

mike1808:
Thanks @tlwr .

Can you please run tests by running `./bin/test? Because I see you have syntax errors (your newly created function requires two arguments but you pass one)

tlwr:
I've force pushed with the correct number of args, and inverted the ordering because `addRouterErrorHeader(rw, "unknown_route")` seems better than the reverse.

Sorry for being lazy about running the tests - getting routing-release to be able to run the tests is a pain. I'm about to log off for tonight but I'll have a crack at cloning it in the morning

mike1808:
@tlwr to run tets:

1. clone routing-release, call `./scripts/update` inside of it
2. call `./scripts/start-docker-for-testing.sh`
3. navigate to `/routing-release/src/code.cloudfoundry.org/gorouter` inside Docker container and run `./bin/test`

tlwr:
Thanks @mike1808 - tests have run:

Unit changed:
```
+ ginkgo -keepGoing -trace -progress -r -failOnPending -randomizeAllSpecs -p -race -nodes 2 handlers/
Running Suite: Handlers Suite
=============================
Random Seed: 1570491918 - Will randomize all specs
Will run 143 specs

Running in parallel across 2 nodes

•••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••••
Ran 143 of 143 Specs in 1.149 seconds
SUCCESS! -- 143 Passed | 0 Failed | 0 Pending | 0 Skipped
```

I've run them before but they take 7GB of disk space and they make my poor macbook air cry, so I usually wait until I'm home and can run them on my linux machine :D

I also ran all of the tests except the Crypto tests. The latest commit changes the proxy tests with the updated handler behaviour (404 -> 503).

I had an integration test flake on me, but not reproducibly, nor when I ran with `ginkgo -nodes 1`

handlers/helpers.go
```diff
@@ -5,10 +5,26 @@ import (
...
 	"code.cloudfoundry.org/gorouter/logger"
 	"github.com/uber-go/zap"
 )
 
+const (
+	cacheMaxAgeSeconds = 2
+)
+
+func addRouterErrorHeader(rw http.ResponseWriter, val string) {
+	rw.Header().Set(router_http.CfRouterError, val)
```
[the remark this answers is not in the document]
tlwr: Ok, I have the following actions:

* use the helper function for setting `X-Cf-RouterError` everywhere (`handlers/routeservice.go` and `proxy/handler/request_handler.go`)
* lowercase the `Sets` in `It("Sets X-Cf-RouterError to unknown_route", func() {` in `handlers/lookup_test.go`.
* squash the three commits into a single commit

Sound good?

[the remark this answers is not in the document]
mike1808: 1) use it everywhere - 👍 
2) lowercase - 👍 
3) squash 👎 make each commit to do one things, e.g.
* the first commit refactors the codebase to use your created function
* the second commit adds new status code and error messages and tests with it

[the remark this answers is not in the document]
tlwr: Hi I've pushed three commits

* Helpers
* When `It` is followed by a verb, it should be lower case
* 404 -> 503

tlwr:
There were conflicts due to the lowercasing, which I have resolved with a rebase

mike1808:
LGTM! Thank you so much for this contribution!

tlwr:
Thanks!

ameowlia:
Hi @tlwr  👋

I was thinking about this problem with @mcwumbly recently and I have some delayed thoughts. 

❓ Did this PR solve the problem you were trying to solve?

By your own problem statement you want to differentiate between a "that route does not exist" (this previously and currently returns a 404) and "there are no backends mapped to that route" (was a 404, now is sometimes a 503 and sometimes a 404). I understand and sympathize with that problem, however this PR does not fix this completely.

### What currently happens

**Scenario 0: curl a route that DNE**
1. Curl meow.DOMAIN, a domain that does not exist, not even in CC API.

```
404 Not Found: Requested route ('dora.magentapink.cf-app.com') does not exist.
```

**Scenario 1: route exists in CC API, but no backends are mapped to it**
1. Push an app with a route
1. Curl it and get a 200
1. Unmap route
1. Curl the route with exists in CAPI, but has no backends and get:

```
404 Not Found: Requested route ('dora.magentapink.cf-app.com') does not exist.
```

**Scenario 2: make gorouter prune route a "bad backend" so there are no backends for a route in the gorouter routing table**

1. Push an app with a route
1. Curl it get a 200
1. Change the iptables rules so it will fail. This is what I did:
```
# remove the dnat rule that sends the traffic from gorouter to the app
iptables -t nat -D netin--d45e27c5-53ac-4719-7b -d 10.0.1.11/32 -p tcp -m tcp --dport 61006 -j DNAT --to-destination 10.255.22.5:61001

# add a dnat rule that sends traffic from gorouter for 10.255.22.5 to 10.255.22.6
# this will cause a failure because gorouter will detect that the app it connected to is not the app it thought it should connect to and it will prune that route and attempt to retry
iptables -t nat -A netin--d45e27c5-53ac-4719-7b -d 10.0.1.11/32 -p tcp -m tcp --dport 61006 -j DNAT --to-destination 10.255.22.4:61001
```
4. Stop the route emitter so it won't keep emitting the route
5. Curl the route once

```
# On the first curl only
$ curl dora.magentapink.cf-app.com
503 Service Unavailable

# This is where gorouter removes the bad backend from its route table
```
6. Curl again (and again and again for the next 1-10ish seconds)
```
$ curl dora.magentapink.cf-app.com
503 Service Unavailable: Requested route ('dora.magentapink.cf-app.com') has no available endpoints.

# Eventually gorouter goes through and prunes it's routes table. It will remove route entries with no backends in them.
```

7. Curl again and forever after
```
$ curl dora.magentapink.cf-app.com
404 Not Found: Requested route ('dora.magentapink.cf-app.com') does not exist.
```

If we didn't stop the route emitter, it would cycle through all of these results as the bad route continues to be added, removed, and pruned form the gorouter route table.


### Thoughts 

* To me, this seems even more confusing than the previous state when we returned a 404 instead of a 503. I feel like this 503 is leaking gorouter logic to the end user. Now the end user gets a different response when (1) the backend has been removed from the route, but gorouter has not pruned the empty route and (2) when the gorouter has pruned the empty route. 

* In order to get a 404 when the route DNE exist in the CC API or Gorouter and a 503 when the route does exist, but there are no backends would require the gorouter to get routes from the CC API. This would be a very large change to the system.

* I don't have a fix or solution in mind, this just piqued my interest since @mcwumbly and I were digging around in this part of the code.

* We brought up this ask (differentiating between no route and a route with no backends) with the cf-for-k8s networking PM so that they can architect better in the newer project from the get go.

### Questions

My main questions are: 
❓ Did this PR solve the problem you were trying to solve?
❓ Do you rely on this 503 in any scripting?

tlwr:
I can't remember my exact thought process but I remember deliberating a lot over the status codes. The 503 / 404 was a compromise and I think my original logic of

- `When the pool is empty, return a 521 Web Server Is Down`
- `When a route is not found, return a 523 Origin Is Unreachable`

is more precise, although suffers from the challenges you outline above.

---

We originally started looking into this change because when there are no instances in the pool, it can be a platform issue (eg mTLS problems), and so we wanted it not to be a 404 (which is a very noisy status code), so that we could alert on these sorts of problems. I'm still sure that 404 is not the best status code

---

The current behaviour aside, the most useful behaviour is the contents of the `X-Cf-RouterError` header (eg `endpoint_failure`), regardless of the HTTP status code.

Now that CF is using tags in logs and metrics (and Prometheus) in some places. If we expose a metric with the reason for the error in a tag/label called `reason` (for example):

```
response_errors{reason="endpoint_failure"} 10
response_errors{reason="invalid_cf_app_instance_header"} 3
response_errors{reason="empty_host"} 5
```

These metrics are much more useful to us than the status codes (which end up being bucketed into 4xx/5xx anyway), as they are unique, more descriptive, and the desire (above) to stick to more well-known status codes. This would decouple the current gorouter behaviour from the Istio wrangling that (I presume) y'all are having to do, and be better for for observing gorouter.

I'm willing to have a crack at implementing the above behaviour

PR #257, opened 2019-10-25
Lookup handler should not expose LB IP

### LINKED ISSUE #256: Private IP exposure of ELB when making http1.0 requests
Thanks for submitting an issue to `gorouter`. We are always trying to improve! To help us, please fill out the following template.

## Issue

[provide quick introduction so this issue can be triaged]
When a request is made which uses http1.0 with an empty host header, it seems to expose the internal IP address of the ELB.

[tell us about how this impacts your users or production or dev environments, to help us prioritize]
This is a private IP disclosure and doesn't otherwise impact our customers.

## Context

[provide more detailed introduction]
We use v12.0.0 cf-deployment and our environment is behind classic AWS ELB's.

## Steps to Reproduce

[ordered list the process to finding and recreating the issue, example below]
A simple curl is enough to expose the internal IP of the elb:

`$ curl --http1.0 https://foo.cloud.pcftest.com -H 'Host: '`
`404 Not Found: Requested route ('10.10.2.218') does not exist.`

## Expected result

[describe what you would expect to have resulted from this process]
I would expect not to see the internal IP address in the return result

## Current result

[describe what you you currently experience from this process, and thereby explain the bug]
I see what looks like the internal IP address of the ELB in the 404 return result.

Our routers are on 10.10.16.0/20 and 10.10.80.0/20 networks. Our ELB's are on the 10.10.2.0/24 network, which is why I think this request is exposing an ELB IP address.

## Possible Fix

[not obligatory, but suggest fixes or reasons for the bug]

## `name of issue` Output Results

[if relevant, include a screenshot]

tlwr:
A short explanation of the proposed change
---

gorouter requires the Host header to know to which backend to proxy to.

The Host header is optional in HTTP/1.0, so we must explicitly check
that the Host is set.

Also, when some load balancers which are used to load balance gorouters (e.g. AWS Classic ELB) receive HTTP/1.0 requests without a host headers, they optimistically add their IP address to the host header.

Therefore to not expose the internal IP address of the load balancer, we must check that the Host is not also the source of the request.

It is not valid for the Host header to be an IP address, because Gorouter should not have an IP address as a route.

See (search "private ip") https://aws.amazon.com/premiumsupport/knowledge-center/elb-fix-failing-health-checks-alb/ for more information about the Host header and ELBs. This should not affect healthchecking because there is a different handler for health.

An explanation of the use cases your change solves
---

Closes #256

Instructions to test the behavior 
---

See #256, in which they reported:

```
curl https://foo.cloud.pcftest.com -H 'Host: ' --http1.0
```

returns:

```
404 Not Found: Requested route ('10.10.2.218') does not exist.
```

---

Gorouter should:

* return a 400 BadRequest with an `empty_host` value for `X-Cf-RouterError`
* record a bad request metric.

Gorouter should do this if:

* the Host header is _actually_ empty
* if the Host header is _presumed_ to be empty before the load balancer fiddled with it

---

```
404 Not Found: Requested route ('10.10.2.218') does not exist.
```

---

N/A

Checklist
---

### DIFF
--- a/handlers/lookup.go
+++ b/handlers/lookup.go
@@ -36,6 +36,26 @@ func NewLookup(registry registry.Registry, rep metrics.ProxyReporter, logger log
 }
 
 func (l *lookupHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
+	// gorouter requires the Host header to know to which backend to proxy to.
+	//
+	// The Host header is optional in HTTP/1.0,
+	// so we must explicitly check that the Host is set.
+	//
+	// Also, when some load balancers which are used to load balance gorouters
+	// (e.g. AWS Classic ELB)
+	// receive HTTP/1.0 requests without a host headers,
+	// they optimistically add their IP address to the host header.
+	//
+	// Therefore to not expose the internal IP address of the load balancer,
+	// we must check that the Host is not also the source of the request.
+	//
+	// It is not vali for the Host header to be an IP address,
+	// because Gorouter should not have an IP address as a route.
+	if r.Host == "" || hostWithoutPort(r.Host) == hostWithoutPort(r.RemoteAddr) {
+		l.handleMissingHost(rw, r)
+		return
+	}
+
 	pool := l.lookup(r)
 	if pool == nil {
 		l.handleMissingRoute(rw, r)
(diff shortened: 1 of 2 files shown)


adobley:
Looks good to me! Thanks! This is great :D I'll merge this and bump the gorouter submodule in routing-release.

PR #262, opened 2020-02-03
Set vcap_id SameSite value based on JSESSIONID

domdom82:

Set the "SameSite" value of the `__VCAP_ID__` cookie based on the value of the `JSESSIONID` cookie
that came back from the app.

Currently, gorouter does not set a value for "SameSite" at all. Some browsers (eg. Chrome, see https://www.chromestatus.com/feature/5088147346030592) used to treat a missing value as "None" which allowed XS sending of the cookie. This default will change to "Lax" which will prevent the cookie from being sent in some use cases:
- A customer runs a CF app that embeds external content (e.g. via an iframe) which requires cookies sent to the main site be forwarded to the embedded site.

Since JSESSIONID and __VCAP_ID__ are tightly coupled, I propose to add the "SameSite" value of JSESSIONID to __VCAP_ID__, just as we do for "secure" and "maxAge".

`curl -v https://my-app-that-sets-JSESSIONID.com`

"curl" output should contain:

```
Set-Cookie: JSESSIONID=abc123; HttpOnly; Secure; SameSite=Lax
Set-Cookie: __VCAP_ID__=c635e2d9-3a06-45ce-54f6-dbe7; Path=/; HttpOnly; Secure; SameSite=Lax
```

"curl" output contains:

```
Set-Cookie: JSESSIONID=abc123; HttpOnly; Secure; SameSite=Lax
Set-Cookie: __VCAP_ID__=c635e2d9-3a06-45ce-54f6-dbe7; Path=/; HttpOnly; Secure
```

### DIFF
--- a/proxy/round_tripper/proxy_round_tripper.go
+++ b/proxy/round_tripper/proxy_round_tripper.go
@@ -294,6 +294,7 @@ func setupStickySession(
 ) {
 	secure := false
 	maxAge := 0
+	sameSite := http.SameSite(0)
 
 	// did the endpoint change?
 	sticky := originalEndpointId != "" && originalEndpointId != endpoint.PrivateInstanceId
@@ -305,6 +306,7 @@ func setupStickySession(
 				maxAge = v.MaxAge
 			}
 			secure = v.Secure
+			sameSite = v.SameSite
 			break
 		}
 	}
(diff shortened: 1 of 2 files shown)


adobley:
Howdy @domdom82! Thanks for the PR and looking out in the #networking CF slack channel. This looks good to us. We'll bump the module in routing-release and get it through our pipelines.

🎉🎉🎉🎉🎉

@adobley & @KauzClay

PR #270, opened 2020-08-06
Add a configurable TLS Handshake timeout when proxying to applications

aminjam:
Fixes: cloudfoundry/routing-release#177

In this PR, we are only adding `TLSHandshakeTimeout` for the `BackendTemplate`. We were not sure if this value makes sense to be added to `RouteServiceTemplate` or not. We've [defaulted this value to 10s](https://github.com/golang/go/blob/master/src/net/http/transport.go#L52), but if you think this needs to be exposed via routing-release, we can send a PR for making it configurable from BOSH. Please let us know if you'd like us to change or update any part of this PR. 

After this PR is merged, the steps introduced in the isse can be used to see the expected value.

* [x] I have run all the unit tests using `scripts/run-unit-tests-in-docker`


proxy/proxy_test.go
```diff
@@ -1031,6 +1031,78 @@ var _ = Describe("Proxy", func() {
...
+				}
+
+				test_util.RegisterAddr(r, "backend-with-tcp-only", nl.Addr().String(), rcfg)
+			})
+
+			AfterEach(func() {
+				nl.Close()
+			})
+
+			Context("when the server cert does not match the client", func() {
```
sunjayBhatia, line 1092: nit: change this context to something a bit more descriptive

proxy/proxy_test.go
```diff
@@ -1031,6 +1031,78 @@ var _ = Describe("Proxy", func() {
 				})
 			})
 		})
+
+		Context("when attempting to establish TLS handshake with a TCP application", func() {
```
sunjayBhatia, line 1035: Can we frame this test (description at least, maybe the test implementation as well?) so it is less specific to a TCP application. Changing the context description to something about how the TLS handshake is not reciprocated by the upstream (and a comment that we are using a TCP server that will not continue the TLS handshake) may be enough, just to make it clear this is not a TCP app only issue

ameowlia: +1

ameowlia:
Thanks for doing this so quickly @aminjam,

Besides the wording things that @sunjayBhatia mentioned, this looks good to me. A 10s default seems fine. I don't think we need to make it bosh configurable.

If you can make the changes mentioned above, then I'll merge it in and let it run through our pipelines. 

Thanks!

PR #278, opened 2021-04-06
Update Router to serve HTTP/2 traffic

Gerg:
Adds a manifest property, `enable_http2`, to enable the Gorouter to accept HTTP/2 connections. The property changes the protocol checking middleware and the `serveHTTPS` function to allow HTTP/2. Note that connections made using HTTP/2 will still be proxied to backends over HTTP/1.1 because Envoy isn't currently configured to accept HTTP/2.

We added integration tests to check HTTP/2 (and updated the old ones to check that gorouter can still serve HTTP/1.1). However, we weren't able to get a router/router test working because making HTTP/2 requests to the test Server resulted in broken pipes. We believe this is due to differences in the test Server setup and the `main.go` setup used in the integration tests. This will require additional investigation.

[cloudfoundry/routing-release#200]

### DIFF
--- a/integration/main_test.go
+++ b/integration/main_test.go
@@ -255,12 +255,14 @@ var _ = Describe("Router Integration", func() {
 			clientTLSConfig *tls.Config
 			mbusClient      *nats.Conn
 		)
+
 		BeforeEach(func() {
 			cfg, clientTLSConfig = createSSLConfig(statusPort, proxyPort, sslPort, natsPort)
-
 		})
+
 		JustBeforeEach(func() {
 			var err error
+			cfg.EnableHTTP2 = true
 			writeConfig(cfg, cfgFile)
 			mbusClient, err = newMessageBus(cfg)
 			Expect(err).ToNot(HaveOccurred())
@@ -297,7 +299,6 @@ var _ = Describe("Router Integration", func() {
 	})
 
 	Context("Drain", func() {
-
 		BeforeEach(func() {
 			cfg = createConfig(statusPort, proxyPort, cfgFile, defaultPruneInterval, defaultPruneThreshold, 1, false, 0, natsPort)
 		})
--- a/integration/main_test.go
+++ b/integration/main_test.go
@@ -20,6 +20,8 @@ import (
 	"syscall"
 	"time"
 
+	"golang.org/x/net/http2"
+
 	tls_helpers "code.cloudfoundry.org/cf-routing-test-helpers/tls"
 	"code.cloudfoundry.org/gorouter/config"
 	"code.cloudfoundry.org/gorouter/mbus"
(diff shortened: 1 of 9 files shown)


integration/main_test.go
```diff
@@ -247,9 +249,107 @@ var _ = Describe("Router Integration", func() {
...
+				for {
+					select {
+					case <-runningTicker.C:
+						runningApp1.Register()
+					case <-done:
+						return
+					}
+				}
+			}()
+			Eventually(func() bool { return appRegistered(routesUri, runningApp1) }).Should(BeTrue())
```
ameowlia, line 343: Missing test: what happens if you use the non-h2 client here? 

Client from the test above:
```
client := &http.Client{Transport: &http.Transport{TLSClientConfig: clientTLSConfig}}
```

router/router_test.go
```diff
@@ -1350,6 +1350,50 @@ var _ = Describe("Router", func() {
 
 	})
 
+	Context("serving multiple http versions", func() {
```
ameowlia, line 1353: This context says "serving multiple http versions" but then the only test is about HTTP/1.1 and I don't see any references to other versions.

This the fact that the h2 test is missing related to your comment in the PR: 
> We were not able to get a router/router test working because making
HTTP/2 requests to the test Server resulted in broken pipes. We believe
this is due to differences in the test Server setup and the main.go
setup used in the integration tests. This will require additional
investigation.

❓ So are you missing test coverage here in order to keep the tests passing? 
❓ If so, can you add the test even if it is broken so others can more easily look into it?
❓ Can you also add more details about how the test is failing?

router/router.go
```diff
@@ -236,6 +236,10 @@ func (r *Router) serveHTTPS(server *http.Server, errChan chan error) error {
 		ClientAuth:   r.config.ClientCertificateValidation,
 	}
 
+	if r.config.EnableHTTP2 {
+		tlsConfig.NextProtos = []string{"h2"}
```
ameowlia, line 240: ❓ Can you explain what this does?

ameowlia (changes requested):

integration/main_test.go
ctlong: It should succeed. Just updated the tests to reflect that.

router/router_test.go
mkocher: The challenge that we ran into when trying to write the test is that the test setup for router_test.go doesn't work with h2 as the test setup is not using the production setup code and it wasn't clear how to change it. This is probably because h2 requires TLS.

We added test coverage for all these cases in the integration tests, so I think it makes sense to just delete this one loner test.

router/router.go
ctlong: Adding `h2` to the TLS config's list of available protocols lets the tlsListener try to negotiate HTTP/2 via ALPN. From the `Serve` function documentation:

> HTTP/2 support is only enabled if the Listener returns *tls.Conn connections and they were configured with "h2" in the TLS Config.NextProtos. - https://golang.org/pkg/net/http/#Serve

mkocher:
@ameowlia  is this good to go?

ctlong:
@ameowlia is this PR good to go?

ameowlia:
Apologizes for the delay. Github comments are not the best way to ping me 😅 

Acceptance Steps
* Set enable_http2 to true
* Deployed this code
* Built [this h2 go client](https://gist.github.com/ameowlia/03d13315135baf9bf1ff3b5a09bb98da) with GOOS=linux
* `bosh scp`ed it to the router VM
* Sent h2 traffic directly to the gorouter for an app
* Saw the response was h2
* Set enable_http2 to false and redeploy
* Sent h2 traffic directly to the gorouter for an app
* Saw it fail to connect with h2

PR #282, opened 2021-05-14
reverse proxy retruns 502 instead of 499 when client disconnects

randomtask1155:

Recently observed the gorouter will randomly return a 499 or 502 status code when the originating client disconnects   ([tracker story](https://www.pivotaltracker.com/n/projects/2477027/stories/177007392)).

The most common case for this to happen is when gorouter reverse proxy is reading from downstream client who originated the request and the TCP socket is closed abruptly.  

The 502 and 499 status codes have a specific meaning and it can make troubleshooting difficult when they are incorrectly reported.  When we see a 502 error we expect the backend has failed in some way vs when we see a 499 where we expect the frontend to have failed.   It is challenging for operation teams to determine the severity of the situation when these results are misreported. 

There are a few OSS projects that solve this very same issue by simply checking the originating client context before passing the reverse proxy error through.  Essentially when there is a failure on the reverse proxy we are checking to make sure the frontend is still there, and if not we return a 499.
https://github.com/zalando/skipper/issues/1631
https://github.com/grafana/metrictank/pull/1821/files

to reproduce simply curl and app through gorouter that uploads a large file.  Then kill the curl command after a second.  You will start to see an even distribution of 499 and 502 errors.  ussing this [app](https://github.com/randomtask1155/simple-http) as an example.

```
#!/bin/bash

for i in {1..1000}
do
	curl -X POST "https://simple-http.appps.domain/post/data" -k -F 'file=@/Users/user/Downloads/ubuntu.iso' &
	sleep 1
	killall curl
done
```

To confirm this fix does not breaking some of the other types of workflow failures i ran 4 different tests for the fix and the results are [here](https://docs.google.com/spreadsheets/d/1bKJ9TdKSyyaGOonwMhMm-35bULNPdSHFTkGgueoaW4s/edit#gid=0)

download kill backend - Run a curl to the backend that downloads a file and use tcpkill to kill the apps tls port on the diego cell
download kill frontend - Run a curl to the backend that downloads a file and kill the curl command in the middle of the download
upload kill backend - same as download kill backend test except we upload a file
upload kill frontend - same as download kill frontend test except we upload a file

expect gorouter to return 499 http status error when originating client abruptly goes away

gorouter will randomly return a 499 or 502 when downstream client cancels request.

i might need to sign the updated CLA.

yes i ran the unit test.  But i had to modify the test script to install golang 1.15 because the ci pipeline image uses 1.16 which was not compatible.  the round tripper code base passed but there were some other errors related to certs that I don't think correlate with this change. 

no

no

### DIFF
--- a/proxy/proxy_test.go
+++ b/proxy/proxy_test.go
@@ -1,6 +1,7 @@
 package proxy_test
 
 import (
+	"bufio"
 	"bytes"
 	"crypto/tls"
 	"crypto/x509"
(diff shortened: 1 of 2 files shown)


ameowlia:
Thanks for submitting this @randomtask1155 ✨ 

Looks reasonable to me (without having tested it). I suggest that whoever picks up the story tests it by running the reproduction steps that @randomtask1155 provided before merging.

jrussett:
Hey @randomtask1155,

Firstly, thanks for submitting this PR!

## Reproduction Woes
I'm having trouble replicating your reproduction steps. I pushed up your sample app and then used the latest ubuntu server iso, but instead of `502`s or `499`s (after I deployed your changes), I receive a ton of `503`s and `404`s:

```
# 10 iterations
$ ./curl-simple-http.sh
curl: no process found
curl: no process found
503 Service Unavailable
curl: no process found
503 Service Unavailable: Requested route ('simple-http.apps.indianred.cf-app.com') has no available endpoints.
curl: no process found
503 Service Unavailable: Requested route ('simple-http.apps.indianred.cf-app.com') has no available endpoints.
curl: no process found
503 Service Unavailable: Requested route ('simple-http.apps.indianred.cf-app.com') has no available endpoints.
curl: no process found
503 Service Unavailable: Requested route ('simple-http.apps.indianred.cf-app.com') has no available endpoints.
curl: no process found
404 Not Found: Requested route ('simple-http.apps.indianred.cf-app.com') does not exist.
curl: no process found
404 Not Found: Requested route ('simple-http.apps.indianred.cf-app.com') does not exist.
curl: no process found
404 Not Found: Requested route ('simple-http.apps.indianred.cf-app.com') does not exist.
curl: no process found
404 Not Found: Requested route ('simple-http.apps.indianred.cf-app.com') does not exist.
curl: no process found
```

I thought that this may be a scale issue, so I scaled the number of application instances to `10` and tried again:
```
curl: no process found
curl: no process found
curl: no process found
curl: no process found
curl: no process found
curl: no process found
```

I then tried making the diego cell larger and used `10` instances of our [proxy app](https://github.com/cloudfoundry/cf-networking-release/tree/develop/src/example-apps/proxy) to upload a `1GB` binary file to the [`/upload` endpoint](https://github.com/cloudfoundry/cf-networking-release/tree/develop/src/example-apps/proxy#upload). The results are the same:
```
$ ./curl-potato-proxy.sh
curl: no process found
curl: no process found
curl: no process found
curl: no process found
curl: no process found
curl: no process found
curl: no process found
curl: no process found
curl: no process found
curl: no process found
curl: no process found
```

## Questions
❓ How many application instances of your sample app did you deploy?
❓ What is the scale of the environment? Size of diego cells, number of cells, etc...
❓ Is there anything else I should know about the setup of the env to reproduce this correctly?

Thanks!
@jrussett

jrussett:
I also noticed that there are no tests for this change. @randomtask1155 could you please add a test to demonstrate and exercise this behavior.

randomtask1155:
i launched a new environment and ran the test again. was able to reproduce it instantly.  I am using an HAPROXY as the load balancer and sometimes depending on the LB config it could cushion the gorouter from the abrupt disconnects.  For example if the LB is connection pooling to the gorouters. 

a more direct way to test this is to bypass the LB. 

```
#!/bin/bash

for i in {1..1000}
do
	curl -X POST "https://192.168.2.10/post/data" -H "Host: simple-http.apps.domain" -k -F 'file=@/Users/user/Downloads/ubuntu-20.04-desktop-amd64.iso' &
	sleep 1
	killall curl
done
```

then tail the gorouters access log

```
:~# tail -f /var/vcap/sys/log/gorouter/access.log  | egrep simple-http
```


i will work on getting a unit test case this week.

randomtask1155:
added test case

swagataroy:
Hello @jrussett I noticed you had some issues. I was testing this today (with @pivotal-Josh-Gainey). We also noticed the 503s and the 404s while tailing the access log. 
Scaling the memory from 128M to 1024M and bypassing the HAProxy helped to reproduce the issue (as in 499s and 502s only) 
Added the following in the /etc/hosts file to bypass HAProxy

`<IP of GoRouter> simple-http.cfapps-23.slot-59.pez.vmware.com`
_tail -f /var/vcap/sys/log/gorouter/access.log  | egrep simple-http_

Do you think you can test and merge the PR? Thanks!

ameowlia:
Hi everyone, 

@Benjamintf1 and I took a look at this again today. 

We ran into a similar issue as @jrussett when trying to reproduce. It turned out to be because the app was running out of memory when we were trying to upload such a large file. Once we increased the app's memory we were able to successfully reproduce the before and after states as you described them. 

@Benjamintf1 and I are happy to merge this in, except that you don't have a CLA signed. 

👉  @randomtask1155 - can you please sign the CLA so we can move forward with this? 

Thanks, 
Amelia and @Benjamintf1

Benjamintf1 (approved):

ameowlia:
🎉🎉🎉🎉🎉🎉🎉

🙏 Thank you so much for continuing to dig into this and submitting this PR @randomtask1155!

ywei2017:
@randomtask1155 thanks for the contribution. I face the same issue. Now we just need to wait for the release.

PR #284, opened 2021-05-20
Improve cross-endpoint request retries

peterellisjones:

---

This PR updates gorouter behaviour so that existing retry logic _to the same endpoint_ for idempotent requests is also attempted _accross endpoints_, for the case when the server fails to provide an HTTP response.

The Go standard library default HTTP transport retries requests on encountering certain errors where the request qualifies as being idempotent. [Specifically](https://github.com/golang/go/blob/master/src/net/http/transport.go#L87-L94):

```
// Transport only retries a request upon encountering a network error
// if the request is idempotent and either has no body or has its
// Request.GetBody defined. HTTP requests are considered idempotent if
// they have HTTP methods GET, HEAD, OPTIONS, or TRACE; or if their
// Header map contains an "Idempotency-Key" or "X-Idempotency-Key"
// entry. If the idempotency key value is a zero-length slice, the
// request is treated as idempotent but the header is not sent on the
// wire.
```

This was implemented in [2013](https://github.com/golang/go/commit/5dd372bd1e70949a432d9b7b8b021d13abf584d1) and updated in [2018](https://go-review.googlesource.com/c/go/+/147457/ ) to add support for idempotency key headers. 

Errors which allow retries to take place are an [HTTP2 cached connections issue](https://github.com/golang/go/blob/master/src/net/http/transport.go#L681) and the following [two errors](https://github.com/golang/go/blob/master/src/net/http/transport.go#L713-L723):

```golang
if _, ok := err.(transportReadFromServerError); ok {
  // We got some non-EOF net.Conn.Read failure reading
  // the 1st response byte from the server.
  return true
}
if err == errServerClosedIdle {
  // The server replied with io.EOF while we were trying to
  // read the response. Probably an unfortunately keep-alive
  // timeout, just as the client was writing a request.
  return true
}
```

Retrying when the above two errors occur runs the risk of [mutating non-idempotent requests](https://github.com/cloudfoundry/gorouter/blob/379860daa83a162ffe0b6039eafb7c8bfa1eaccf/proxy/fails/classifier_group.go#L11) causing bad things to happen (eg duplicate charges), which is why the go standard library only attempts to make these requests when the request meet the idempotency criteria. The current Gorouter ProxyRoundTripper delegates backend requests to the dropsondeRoundTripper which itself delegates requests to the standard library default transport. By virtue of delegating to transport.go's RoundTrip method the Gorouter _already_ retries and accepts the risk of retrying potentially mutating non-idempotent requests, when those requests meet the idempotency criteria required by [http.Request.isReplayable()](https://github.com/golang/go/blob/5c489514bc5e61ad9b5b07bd7d8ec65d66a0512a/src/net/http/request.go#L1413-L1427). As this risk is already accepted and acted on by the Gorouter this PR updates behaviour so that:
1. empty server response errors (ie `io.EOF` errors) for qualifying requests are retried
2. retries happen across endpoints

At least [one scenario](https://github.com/cloudfoundry/routing-release/issues/187) where extending idempotent request retry logic to other endpoints would help improve the stability of Cloud Foundry is when route integrity is enabled and the Envoy sidecar is listening but the application is not yet listening, either because it is being started or has crashed or some other reason. In this situation the following happens:

1. The gorouter successfully establishes a TCP connection to the Envoy TCP Proxy
2. Envoy cannot proxy the request so does not respond*
4. The gorouter proxy round tripper gets an `io.EOF` error from the standard library HTTP transport and returns it to the user as an [`endpoint_failure`](https://docs.cloudfoundry.org/adminguide/troubleshooting-router-error-responses.html#app-errors)-type 502 error.

--- 

I was unable to find an implementation I was completely happy but I went with this: I've added a check in `ProxyRoundTripper#RoundTrip` to replace `io.EOF` with a new error type in cases where the request matches the idempotency criteria. This error is then checked by a new classifier added to the `RetriableClassifiers`  group. This avoids having to refactor the `Classifier` interface to add an `http.Request` argument, which would have needed a lot of refactoring to handle other areas in the code where classifier is called, such as when [pruning endpoints](https://github.com/cloudfoundry/gorouter/blob/main/route/pool.go#L398). If you have suggestions for better implementations, please share!

I implemented the unit tests as a table-based test in order to mirror the tests in the standard library's [transport_test.go](https://github.com/golang/go/blob/master/src/net/http/transport_test.go#L5683-L5734), but with some more cases as there were are a few code paths missing coverage in the standard library tests.

~This PR is not yet mergeable because I'm still waiting for some CLA-related admin, which will take about a week, but I'm publishing now to hopefully start a discussion about this change.~

cheers,

Pete

*After a successful three-way TCP handshake between the gorouter and Envoy, the gorouter sends the HTTP request data. Envoy does not respond so the gorouter closes the TCP connection after about half a second.

### DIFF
--- a/proxy/fails/basic_classifiers.go
+++ b/proxy/fails/basic_classifiers.go
@@ -3,10 +3,13 @@ package fails
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"errors"
 	"net"
 
 	"context"
 )
+
+var IdempotentRequestEOFError = errors.New("EOF (via idempotent request)")
 
 var AttemptedTLSWithNonTLSBackend = ClassifierFunc(func(err error) bool {
 	switch err.(type) {
@@ -70,4 +73,8 @@ var UntrustedCert = ClassifierFunc(func(err error) bool {
 	default:
 		return false
 	}
+})
+
+var IdempotentRequestEOF = ClassifierFunc(func(err error) bool {
+	return err == IdempotentRequestEOFError
 })
--- a/proxy/fails/classifier_group.go
+++ b/proxy/fails/classifier_group.go
@@ -19,6 +19,7 @@ var RetriableClassifiers = ClassifierGroup{
 	RemoteHandshakeTimeout,
 	UntrustedCert,
 	ExpiredOrNotYetValidCertFailure,
+	IdempotentRequestEOF,
 }
 
 var FailableClassifiers = ClassifierGroup{
--- a/proxy/fails/classifier_group_test.go
+++ b/proxy/fails/classifier_group_test.go
@@ -42,6 +42,7 @@ var _ = Describe("ClassifierGroup", func() {
 			Expect(rc.Classify(x509.UnknownAuthorityError{})).To(BeTrue())
 			Expect(rc.Classify(x509.CertificateInvalidError{Reason: x509.Expired})).To(BeTrue())
 			Expect(rc.Classify(errors.New("i'm a potato"))).To(BeFalse())
+			Expect(rc.Classify(fails.IdempotentRequestEOFError)).To(BeTrue())
 		})
 	})
 
@@ -59,6 +60,7 @@ var _ = Describe("ClassifierGroup", func() {
 			Expect(pc.Classify(x509.UnknownAuthorityError{})).To(BeTrue())
 			Expect(pc.Classify(x509.CertificateInvalidError{Reason: x509.Expired})).To(BeTrue())
 			Expect(pc.Classify(errors.New("i'm a potato"))).To(BeFalse())
+			Expect(pc.Classify(fails.IdempotentRequestEOFError)).To(BeTrue())
 		})
 	})
 })
--- a/proxy/round_tripper/proxy_round_tripper.go
+++ b/proxy/round_tripper/proxy_round_tripper.go
@@ -3,6 +3,7 @@ package round_tripper
 import (
 	"context"
 	"errors"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -147,6 +148,12 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
 			res, err = rt.backendRoundTrip(request, endpoint, iter, logger)
 
 			if err != nil {
+				// io.EOF errors are considered safe to retry for certain requests
+				// Replace the error here to track this state when classifying later.
+				if err == io.EOF && isIdempotent(request) {
+					err = fails.IdempotentRequestEOFError
+				}
+
 				iter.EndpointFailed(err)
 				logger.Error("backend-endpoint-failed", zap.Error(err), zap.Int("attempt", retry+1), zap.String("vcap_request_id", request.Header.Get(handlers.VcapRequestIdHeader)))
 
@@ -370,4 +377,22 @@ func requestSentToRouteService(request *http.Request) bool {
 	sigHeader := request.Header.Get(routeservice.HeaderKeySignature)
 	rsUrl := request.Header.Get(routeservice.HeaderKeyForwardedURL)
 	return sigHeader != "" && rsUrl != ""
+}
+
+// Matches behavior of isReplayable() in standard library net/http/request.go
+// https://github.com/golang/go/blob/5c489514bc5e61ad9b5b07bd7d8ec65d66a0512a/src/net/http/request.go
+func isIdempotent(request *http.Request) bool {
+	if request.Body == nil || request.Body == http.NoBody || request.GetBody != nil {
+		switch request.Method {
+		case "GET", "HEAD", "OPTIONS", "TRACE", "":
+			return true
+		}
+		// The Idempotency-Key, while non-standard, is widely used to
+		// mean a POST or other request is idempotent. See
+		// https://golang.org/issue/19943#issuecomment-421092421
+		if request.Header.Get("Idempotency-Key") != "" || request.Header.Get("X-Idempotency-Key") != "" {
+			return true
+		}
+	}
+	return false
 }
--- a/proxy/round_tripper/proxy_round_tripper_test.go
+++ b/proxy/round_tripper/proxy_round_tripper_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"net/http/httptest"
@@ -19,6 +20,7 @@ import (
 	sharedfakes "code.cloudfoundry.org/gorouter/fakes"
 	"code.cloudfoundry.org/gorouter/handlers"
 	"code.cloudfoundry.org/gorouter/metrics/fakes"
+	"code.cloudfoundry.org/gorouter/proxy/fails"
 	errorClassifierFakes "code.cloudfoundry.org/gorouter/proxy/fails/fakes"
 	"code.cloudfoundry.org/gorouter/proxy/handler"
 	"code.cloudfoundry.org/gorouter/proxy/round_tripper"
@@ -28,6 +30,7 @@ import (
 	"code.cloudfoundry.org/gorouter/test_util"
 
 	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/gbytes"
 	"github.com/uber-go/zap"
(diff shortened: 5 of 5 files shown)


peterellisjones:
> * [ ]  I have viewed signed and have submitted the Contributor License Agreement
> * [x]  I have made this pull request to the `main` branch
> * [x]  I have run all the unit tests using `scripts/run-unit-tests-in-docker` from [routing-release](https://github.com/cloudfoundry/routing-release).
> * [ ]  (Optional) I have run Routing Acceptance Tests and Routing Smoke Tests on bosh lite
> * [ ]  (Optional) I have run CF Acceptance Tests on bosh lite
> 
> This PR updates gorouter behaviour so that existing retry logic _to the same endpoint_ for idempotent requests is also attempted _accross endpoints_.
> 
> The Go standard library default HTTP transport retries requests on encountering network errors where the request qualifies as being idempotent. [Specifically](https://github.com/golang/go/blob/master/src/net/http/transport.go#L87-L94):
> 
> ```
> // Transport only retries a request upon encountering a network error
> // if the request is idempotent and either has no body or has its
> // Request.GetBody defined. HTTP requests are considered idempotent if
> // they have HTTP methods GET, HEAD, OPTIONS, or TRACE; or if their
> // Header map contains an "Idempotency-Key" or "X-Idempotency-Key"
> // entry. If the idempotency key value is a zero-length slice, the
> // request is treated as idempotent but the header is not sent on the
> // wire.
> ```
> 
> This was implemented in [2013](https://github.com/golang/go/commit/5dd372bd1e70949a432d9b7b8b021d13abf584d1) and updated in [2018](https://go-review.googlesource.com/c/go/+/147457/) to add support for idempotency key headers.
> 
> The current gorouter ProxyRoundTripper delegates backend requests to the dropsondeRoundTripper which itself delegates requests to the standard library default transport. Therefore requests to a given backend are currently retried using this logic when those requests meet the idempotency criteria. On retrying, if the second request also fails with a network error, that error is escalated to the ProxyRoundTripper as an `io.EOF` error. Although the gorouter does attempt to retry [many classes of errors](https://github.com/cloudfoundry/gorouter/blob/main/proxy/fails/classifier_group.go#L13-L22) against different endpoints, it does not attempt to retry these kind of network failure requests.
> 
> At least [one scenario](https://github.com/cloudfoundry/routing-release/issues/187) where extending idempotent request retry logic to other endpoints would help improve the stability of Cloud Foundry is when route integrity is enabled and the Envoy sidecar is listening but the application is not yet listening, either because it is being started or has crashed or some other reason. In this situation the following happens:
> 
> 1. The gorouter successfully establishes a TCP connection to the Envoy TCP Proxy
> 2. Envoy cannot proxy the request so does not respond*
> 3. If the request met the idempotency criteria it is attempted again against the same endpoint and fails for the same reason
> 4. The gorouter proxy round tripper gets an `io.EOF` error from the standard library HTTP transport and returns it to the user as an [`endpoint_failure`](https://docs.cloudfoundry.org/adminguide/troubleshooting-router-error-responses.html#app-errors)-type 502 error.
> 
> This PR adds a step between steps 3 and 4 which retries qualifying requests against other endpoints.
> 
> I was unable to find an implementation I was completely happy with so I'm presenting the 'least-worst' implementation I could think of: I've added a check in `ProxyRoundTripper#RoundTrip` to replace `io.EOF` with a new error type in cases where the request matches the idempotency criteria. This error is then checked by a new classifier added to the `RetriableClassifiers` group. This avoids having to refactor the `Classifier` interface to add an `http.Request` argument, which would have needed a lot of refactoring to handle other areas in the code where classifier is called, such as when [pruning endpoints](https://github.com/cloudfoundry/gorouter/blob/main/route/pool.go#L398). If you have suggestions for better implementations, please share!
> 
> I implemented the unit tests as a table-based test in order to mirror the tests in the standard library's [transport_test.go](https://github.com/golang/go/blob/master/src/net/http/transport_test.go#L5683-L5734), but with some more cases as there were are a few code paths missing coverage in the standard library tests.
> 
> kind regards,
> 
> Pete

ameowlia:
Hi @peterellisjones , 

Thank you for submitting this PR! I had never heard of the Idempotency-Key before, very cool. Just glancing over the code, it all makes sense to me.

I'll add this to our team's backlog for a more thorough review.

jrussett:
Hi @peterellisjones,

Thank you for taking the time to write all this out and submit a PR. Thanks for the unit tests as well, they're great.

#### Reproducing
The only thing that I'm struggling with is how to test this functionality on an environment.
❓ Do you have steps to manually exercise this functionality? Can you please share them if you do?

Thanks,
@jrussett

peterellisjones:
Hi @jrussett here's how I tested the changes:

### Setup 

Start nats server
```shell
nats-server
```

Start Gorouter 
```shell
cd routing-release/src/code.cloudfoundry.org/gorouter
GOPATH=~/workspace/routing-release GO111MODULE=auto go run main.go
```

Register test apps at  `127.0.0.1:9000` and `127.0.0.1:9001`
```shell
watch -n 10 'nats-pub "router.register" "{ \"host\": \"127.0.0.1\", \"port\": 9000, \"uris\": [ \"test-app.internal\" ], \"app_guid\": \"abc123\" }"'
```

```shell
watch -n 10 'nats-pub "router.register" "{ \"host\": \"127.0.0.1\", \"port\": 9001, \"uris\": [ \"test-app.internal\" ], \"app_guid\": \"abc123\" }"'
 ```
 
I then deployed a simple test app to listening on port 9001 -- this would be the "good" app that I would expect the gorouter to retry requests against when requests fail against a "bad" app running on port 9000

```golang
package main

import (
	"fmt"
	"log"
	"net/http"
	"os"
)

var port = os.Getenv("PORT")
var logger = log.New(os.Stdout, "", 0)

func main() {
	logger.SetFlags(log.Lmicroseconds)

	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		logger.Println("Handling request")
		fmt.Fprintln(w, "OK")
	})

	logger.Printf("Listening on port %s\n", port)
	log.Fatal(http.ListenAndServe(fmt.Sprintf(":%s", port), nil))
}
```
 
 ### Scenario 1: App not listening - no change in behaviour expected
 
When the first app (port 9000) is not running, I verified that all requests are sent to the second app (port 9001), and that no requests fail (the gorouter logs also confirmed this)
 ```
➜ curl  -X GET 127.0.0.1:8081 -H "Host: test-app.internal"
OK
```

This is existing behaviour of the Gorouter, due to [dial](https://github.com/cloudfoundry/gorouter/blob/main/proxy/fails/classifier_group.go#L14) errors being retried. 

### Scenario 2: App fails before sending first byte - retries now expected

I then updated the first app (port 9000) to crash before sending the first byte:

```golang
package main

import (
	"fmt"
	"log"
	"net/http"
	"os"
)

var port = os.Getenv("PORT")
var logger = log.New(os.Stdout, "", 0)

func main() {
	logger.SetFlags(log.Lmicroseconds)

	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		panic("Crash before sending first byte")
	})

	logger.Printf("Listening on port %s\n", port)
	log.Fatal(http.ListenAndServe(fmt.Sprintf(":%s", port), nil))
}
```

In this scenario when the current version of the gorouter is running and a request is sent to the first app, the customer gets a 502 bad gateway error from the gorouter

```shell
curl  -X GET 127.0.0.1:8081 -H "Host: test-app.internal"
```

and the gorouter logs show the EOF error
```
{"log_level":3,"timestamp":1622018215.1270452,"message":"backend-endpoint-failed","source":"gorouter.stdout","data":{"route-endpoint":{"ApplicationId":"","Addr":"127.0.0.1:9000","Tags":null,"RouteServiceUrl":""},"error":"EOF","attempt":1,"vcap_request_id":"964d44ac-03fd-4b2a-4b5f-694a3c6e5bfa"}}
2021/05/26 09:36:55 http: proxy error: EOF
```

I then confirmed that with the gorouter running my changes qualifying requests are always retried against other endpoints and never fail:
```
➜ curl  -X GET 127.0.0.1:8081 -H "Host: test-app.internal"
OK
```

However due to the idempotency criteria, a request that is not a body-less GET/OPTIONS/TRACE/HEAD _or_ does not have X-idempotency-key or idempotency-key headers, still fails with the original behaviour:

```
➜ curl  -X PUT 127.0.0.1:8081 -H "Host: test-app.internal" -d '{}'
```

### Scenario 3: App fails after sending first byte - no change in behaviour expected

The current golang behaviour for http.Transport is that qualifying requests are **not** retried if connections fail _after_ sending the first byte and I wanted to make sure my changes matched this to keep the scope of changes conservative. Therefore to make sure that I wasn't accidentally introducing retries for these requests I created an app that fails after sending a partial response:

```golang
package main

import (
	"fmt"
	"log"
	"net/http"
	"os"
)

var port = os.Getenv("PORT")
var logger = log.New(os.Stdout, "", 0)

func main() {
	logger.SetFlags(log.Lmicroseconds)

	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		w.Write([]byte("OK"))
		if f, ok := w.(http.Flusher); ok {
			f.Flush() // Force write response bytes
		}
		panic("Crashing mid-response")
	})

	logger.Printf("Listening on port %s\n", port)
	log.Fatal(http.ListenAndServe(fmt.Sprintf(":%s", port), nil))
}
```

I then confirmed that in these situations the existing behaviour remains, which is that the transfer is closed mid-stream and _is not retried_
```
➜ curl  -X PUT 127.0.0.1:8081 -H "Host: test-app.internal" -d '{}'
curl: (18) transfer closed with outstanding read data remaining
OK%
```

In this situation the gorouter logs the following, and no retries are attempted.
```
2021/05/26 09:53:42 httputil: ReverseProxy read error during body copy: unexpected EOF
```

### Scenario 4: Envoy listening but app is not - retries now expected

From the perspective of the gorouter this scenario is identical to scenario 2. To verify this I ssh'd into a running Cloud Foundry app to copy an envoy.yaml configuration. I then removed any config I was confident was not relevant, and updated it to listen on port 9000 (replacing the first app), and to proxy traffic to a non-existent app listening on port 7000:

```yaml
static_resources:
  clusters:
    - circuit_breakers:
        thresholds:
          - max_connections: 4294967295
      connect_timeout: 0.250s

      load_assignment:
        cluster_name: 0-service-cluster
        endpoints:
          - lb_endpoints:
              - endpoint:
                  address:
                    socket_address:
                      address: 192.168.1.28 # IP of my laptop
                      port_value: 7000 

      name: 0-service-cluster
      type: STATIC
  listeners:
    - address:
        socket_address:
          address: 0.0.0.0
          port_value: 9000
      filter_chains:
        - filters:
            - name: envoy.tcp_proxy
              typed_config:
                "@type": type.googleapis.com/envoy.extensions.filters.network.tcp_proxy.v3.TcpProxy
                cluster: 0-service-cluster
                stat_prefix: 0-stats
                access_log:
                  - name: envoy.access_loggers.stdout
                    typed_config:
                      "@type": type.googleapis.com/envoy.extensions.access_loggers.stream.v3.StdoutAccessLog
      name: listener-9000
```

I then started Envoy locally with the following Dockerfile
```Dockerfile
FROM envoyproxy/envoy-dev:19210537c7169ed7c8560cdaf16f9b3590db3215
COPY envoy.yaml /etc/envoy/envoy.yaml
RUN chmod go+r /etc/envoy/envoy.yaml
```

And ran it with 
```shell
docker build -t local-envoy .
docker run -it -p 9000:9000 -v $(pwd)/logs:/var/log -e ENVOY_UID=777 local-envoy
```

This gave me the following setup:
* Envoy listening on port 9000 and forwarding traffic to port 7000
* Nothing listening on port 7000 (simulating an app that is not listening because it is restarting or has crashed)
* A healthy app listening on port 9001
* The gorouter configured to send traffic to a single app listening on ports 9000 and 9001

Without the change what happens in this situation is identical to scenario 1: The gorouter gets an empty response from Envoy, the roundtrip call fails with an `io.EOF` error, and the gorouter sends the customer a `502 bad gateway` response.

With the change the failed request to Envoy is successfully retried against the healthy app.

@jrussett, @ameowlia I hope this is useful -- please let me know if you'd like any more details

cheers,

Pete

peterellisjones:
Update: my CLA issues are all resolved (I'm doing this work on behalf of SAP)

ameowlia:
🎉  Thank you @peterellisjones for this PR and for the clear acceptance steps! 🎉

PR #287, opened 2021-07-13
Log more backend endpoint info

46bit:
* A short explanation of the proposed change: log a little more information when backend endpoints fail

* An explanation of the use cases your change solves: we are experiencing a small number of backend failures during platform updates. We would like a little more information in the logs, as provided by this PR

* Instructions to functionally test the behavior change using operator interfaces (BOSH manifest, logs, curl, and metrics): can trigger this log by stopping a traffic-receiving VM with `bosh stop --skip-drain diego-cell/N`

* Expected result after the change: extra context in the logs, like can be seen here and in the PR:`{"log_level":3,"timestamp":"2021-07-13T15:01:30.452379948Z","message":"backend-endpoint-failed","source":"vcap.gorouter","data":{"retriable":true,"num-endpoints":1, [...]}}`

* Current result before the change: the `retriable` and `num-endpoints` fields aren't present, which makes it a little harder to follow what behaviour Gorouter will do next

* [x] I have deployed this in a CF environment and checked the new log fields print as expected

### DIFF
--- a/proxy/round_tripper/proxy_round_tripper.go
+++ b/proxy/round_tripper/proxy_round_tripper.go
@@ -155,9 +155,11 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
 				}
 
 				iter.EndpointFailed(err)
-				logger.Error("backend-endpoint-failed", zap.Error(err), zap.Int("attempt", retry+1), zap.String("vcap_request_id", request.Header.Get(handlers.VcapRequestIdHeader)))
 
-				if rt.retriableClassifier.Classify(err) {
+				retriable := rt.retriableClassifier.Classify(err)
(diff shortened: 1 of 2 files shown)


46bit:
One more thing we'd like to do is make it easier for clients to understand why a request failed. We have a draft commit for that. Any thoughts if you're comfortable leaking [each of these situations[(https://github.com/46bit/gorouter/commit/cbe1c083d288be14885fe9be890c9a650e559826#diff-98906785d04ed7c45fd31c2a9d6d1375afc3fc87ccfcdf42ec7ca3ab3a32013dL31-L38)? https://github.com/46bit/gorouter/commit/cbe1c083d288be14885fe9be890c9a650e559826

ameowlia:
Thank you @46bit for this PR. 

In general, I am very supportive of making cf_router_error header more specific and helpful to endusers. I worry a little bit about leaking security concerns, but I can't think of any specific problems right now.

PR #294, opened 2021-08-11
Add error message to CfRouterError for better debugging

thomas-kaltenbach:
Based on the Idea of @46bit and the [Draft Commit](https://github.com/46bit/gorouter/commit/cbe1c083d288be14885fe9be890c9a650e559826). I took over the topic and created this PR.

Add the error message to the header X-Cf-RouterError.

This change enables the CF Users to debug better their application in case of 502 status codes. In the most cases it is very helpful to see the error message if an `endpoint_failure` occurs. Currently, when the CF User doesn't know why this happens they usually opens a ticket for help. We want to prevent this and add the error message to the header `X-Cf-RouterError`. 
Also @domdom82 provided [here](https://github.com/46bit/gorouter/commit/cbe1c083d288be14885fe9be890c9a650e559826#commitcomment-53572511) some motivation for this PR. 

Provoke an app crash and check the returning header `X-Cf-RouterError`.

Header should be i.e. `endpoint_failure (EOF)` or `endpoint_failure (http: server closed idle connection)`. 
This message can be also found at the accesslog property "RouterError"

Only `endpoint_failure` without error message

### DIFF
--- a/proxy/proxy_test.go
+++ b/proxy/proxy_test.go
@@ -1504,7 +1504,7 @@ var _ = Describe("Proxy", func() {
 
 			resp, body := conn.ReadResponse()
 			Expect(resp.StatusCode).To(Equal(http.StatusBadGateway))
-			Expect(resp.Header.Get("X-Cf-RouterError")).To(Equal("endpoint_failure"))
+			Expect(resp.Header.Get("X-Cf-RouterError")).To(Equal("endpoint_failure (http: server closed idle connection)"))
 			Expect(body).To(Equal("502 Bad Gateway: Registered endpoint failed to handle the request.\n"))
 		})
 
--- a/proxy/round_tripper/error_handler.go
+++ b/proxy/round_tripper/error_handler.go
@@ -1,6 +1,7 @@
 package round_tripper
 
 import (
+	"fmt"
 	"net/http"
 
 	router_http "code.cloudfoundry.org/gorouter/common/http"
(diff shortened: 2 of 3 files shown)


ameowlia:
Hi @thomas-kaltenbach, 


Thank you so much for submitting this PR. I'm sure it will help many future debuggers. 

However, the following unit test is failing. This test was added regarding [this](https://github.com/cloudfoundry/gorouter/pull/284) PR.

```
• Failure [0.002 seconds]
Proxy
/routing-release/src/code.cloudfoundry.org/gorouter/proxy/proxy_test.go:39
  Error Responses
  /routing-release/src/code.cloudfoundry.org/gorouter/proxy/proxy_test.go:1352
    responds to misbehaving host with 502 [It]
    /routing-release/src/code.cloudfoundry.org/gorouter/proxy/proxy_test.go:1388

    Expected
        <string>: endpoint_failure (EOF (via idempotent request))
    to equal
        <string>: endpoint_failure (http: server closed idle connection)

    /routing-release/src/code.cloudfoundry.org/gorouter/proxy/proxy_test.go:1401
```

❓ Can you make sure this unit test passes?

Besides that everything else looks good. 

Thanks!
@ameowlia and @jrussett 🐈  🐈‍⬛

thomas-kaltenbach:
Hi @ameowlia,
thank you for reviewing the PR.
I run the test and rebased my changes to the current main. Unfortunately I couldn't see the same failure you reported.
But I saw that another's were failing because of `invalid memory address or nil pointer dereference`, because in some unit tests the method `err.Error()` was not available. I could solve it by changing the line to `msg = fmt.Sprintf("%s (%s)", msg, err)`.
So, the test run should be now successful.

geofffranks:
Thanks @thomas-kaltenbach! Not sure if you addressed the error above, or if it was a flake, since things seem to work when i test locally now. Going to merge and see what CI says

PR #296, opened 2021-08-20
Cause gorouter to fail to start if it cannot listen on its StatusPort

geofffranks:
[#178665581]

Cause the gorouter process to exit with an error if it's StatusPort is bound by another process, so that the monit status reflects a state similar to what load balancers using the health check reflect (that the process is down).

In cases where something else has bound to port :8080, the new gorouter process will fail to listen, but continue starting up. This could potentially lead to a deployment succeeding in a way that renders all gorouter instances unavailable from the loadbalancer

1. Deploy against a cf-deployment environment
2. SSH into a router instance
3. monit stop gorouter
4. nc -ldk 8080 &
5. monit start gorouter
6. Watch the gorouter fail to start up, with logs emitted indicating that port 8080 is in use

Gorouter will exit on startup and emit a log message for the failure

Gorouter starts up, but does not listen on the StatusPort

### DIFF
--- a/common/component.go
+++ b/common/component.go
@@ -161,8 +161,7 @@ func (c *VcapComponent) Start() error {
 
 	procStat = NewProcessStatus()
 
-	c.ListenAndServe()
-	return nil
+	return c.ListenAndServe()
 }
 
 func (c *VcapComponent) Register(mbusClient *nats.Conn) error {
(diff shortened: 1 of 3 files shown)


reneighbor:
+1

PR #298, opened 2021-09-29
Prune endpoints when for failed route-services

geofffranks:

The routeservice handler will now prune the 0-indexed endpoint from
a request's endpointpool, allowing other endpoints to be used for
the route_service_url until one succeeds, or there are is only one
endpoint left. Any valid endpoints will be recreated as route_emitter
emits them periodically.

When a route-service fails, there is a chance the endpoint
at index 0 of an endpointpool is stale, and has an out of date
route_service_url. Since this is the endpoint used for all of a pool's
route_service_url alls, all calls will fail, even if valid endpoints
have the correct route service URL. When this happens, no stale routes
can be pruned, since none of the endpoints are ever tried (the route
service call always fails first).

Stale routes with bad route-service-urls will be cleaned up

A bad route-service-url on a stale-route in index 0 will result in failures for all requests and never be pruned

* [ x ] I have viewed signed and have submitted the Contributor License Agreement

* [ x ] I have made this pull request to the `main` branch

* [ x ] I have run all the unit tests using `scripts/run-unit-tests-in-docker` from [routing-release](https://github.com/cloudfoundry/routing-release).


geofffranks:
@cphi-vmw FYI

PR #303, opened 2021-12-14
Increment endpoint connection count correctly

aminjam:
### Fix
- Change request_handler to decrement counter after the request has been returned to the client. This fixes an issue for the endpoint connection limits for a WebSocket request.  
- Endpoints should be compared by their fields as oppose to comparing pointers. This will make sure that registry is not updating Stats object for endpoints when it's not necessary.
- Empty pool should now return false for Overloaded

#### Note
I was surprised to see that we don't have much coverage for happy paths for [request_handler_test package](https://github.com/cloudfoundry/gorouter/blob/65207dc5dafa659d343975216701fb2e29e3992a/proxy/handler/request_handler_test.go). This made it difficult to verify the fix for the issue in our unit tests.  I also wasn't able to find any integration tests that would verify this part of the codebase. I verified this change manually by following the steps below. 

### Steps to reproduce
1. Grab a cf-deployment environment
2. Deploy a websocket app (https://github.com/gorilla/websocket/tree/master/examples/chat seems to be reasonable), scale it to 2GB of memory, and only one instance
3. Ensure the gorouter property for router.backends.max_conns is set to 1
4. Use a websocket client to connect to the app (https://github.com/oliver006/ws-client)
5. curl the application from a separate shell while the websocket is still open

### Expected

You should receive a 503 indicating connection limits were exceeded for the endpoint

### Actual

The request succeeds.

### Helpful links:

Websocket app: https://github.com/gorilla/websocket/tree/master/examples/chat
Websocket client: https://github.com/oliver006/ws-client

We ran the websockets in a for loop like this:

```
for i in $(seq 1 2000); do
  ./gen-output | ./ws-client ws://<app>.domain/ws &
done
```

gen-output script (needed to keep the ws-client from stopping in the background do to lack of stdin):
```
#!/bin/bash

while sleep 1; do
  echo sleeping
done
```

### DIFF
--- a/proxy/handler/request_handler.go
+++ b/proxy/handler/request_handler.go
@@ -205,9 +205,11 @@ func (h *RequestHandler) serveTcp(
 			backendConnection, err = net.DialTimeout("tcp", endpoint.CanonicalAddr(), h.endpointDialTimeout)
 		}
 
-		iter.PostRequest(endpoint)
 		if err == nil {
+			defer iter.PostRequest(endpoint)
 			break
+		} else {
+			iter.PostRequest(endpoint)
 		}
 
 		iter.EndpointFailed(err)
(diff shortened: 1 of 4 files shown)


ameowlia (approved):
LGTM

ameowlia:
@aminjam can you get your CLA fixed before I merge this?

PR #304, opened 2022-01-03
Let zipkip handler adhere to w3 spec

stijnvet:

We run our own custom API gateway (based on spring-cloud-gateway) as a downstream service of the gorouter. Therefore, our software redirects all incoming requests to downstream services. So we pass along x-b3 headers (via spring-cloud-sleuth-zipkin) to our downstream services. With the latest change, we noticed that we can no longer correlate gorouter requests with outgoing requests to downsteam services, since the incoming x-b3 headers are incorrect. 

This change will generating trace and span headers according to the w3 standard. With the previous update, the span-id was generated as a string of 32 characters, but according to the standard, it should be of a length of 16 characters. Currently, traces and spans generated by the gorouter will not be able to be picked up by downstream services using libraries like zipkin for java, since it's invalid.

Continue to trace requests coming from the gorouter to downstream services.

Start a gorouter instance with a downstream service running (for example anything that can print request headers). Do a curl via the gorouter to this service and see that the value of the x-b3-spanid header is 32 characters in length.
If the downstream service is running anything with the zipkin library, notice that outgoing requests of the downstream service will have different trace ids.

a x-b3-spanid header with a random 16 character value.

a x-b3-spanid header with a random 32 character value.

### DIFF
--- a/handlers/zipkin.go
+++ b/handlers/zipkin.go
@@ -1,6 +1,7 @@
 package handlers
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/openzipkin/zipkin-go/idgenerator"
@@ -45,11 +46,13 @@ func (z *Zipkin) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.Ha
 	existingTraceID := r.Header.Get(b3.TraceID)
 	existingSpanID := r.Header.Get(b3.SpanID)
 	if existingTraceID == "" || existingSpanID == "" {
-		traceID := idgenerator.NewRandom128().TraceID().String()
+		trace := idgenerator.NewRandom128().TraceID()
 
-		r.Header.Set(b3.TraceID, traceID)
-		r.Header.Set(b3.SpanID, traceID)
-		r.Header.Set(b3.Context, traceID+"-"+traceID)
+		span := fmt.Sprintf("%016x", trace.Low)
+
+		r.Header.Set(b3.TraceID, trace.String())
+		r.Header.Set(b3.SpanID, span)
+		r.Header.Set(b3.Context,  trace.String()+"-"+span)
 	} else {
 		sc, err := b3.ParseHeaders(
 			existingTraceID,
--- a/handlers/zipkin.go
+++ b/handlers/zipkin.go
@@ -1,7 +1,6 @@
 package handlers
 
 import (
-	"fmt"
 	"net/http"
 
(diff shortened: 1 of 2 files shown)


handlers/zipkin.go
```diff
@@ -45,11 +46,13 @@ func (z *Zipkin) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.Ha
 	existingTraceID := r.Header.Get(b3.TraceID)
 	existingSpanID := r.Header.Get(b3.SpanID)
 	if existingTraceID == "" || existingSpanID == "" {
-		traceID := idgenerator.NewRandom128().TraceID().String()
+		trace := idgenerator.NewRandom128().TraceID()
 
-		r.Header.Set(b3.TraceID, traceID)
-		r.Header.Set(b3.SpanID, traceID)
-		r.Header.Set(b3.Context, traceID+"-"+traceID)
+		span := fmt.Sprintf("%016x", trace.Low)
```
seankellner, line 51: Would `idgenerator.NewRandom128().SpanID(trace).String()` be preferable (or even reusing NewRandom128())?

handlers/zipkin_test.go
```diff
@@ -1,6 +1,7 @@
 package handlers_test
 
 import (
+	"fmt"
```
MarcPaquette, line 4: Looks like this import can be dropped?

MarcPaquette:
Other than that errant `fmt` that's an easy fix, this PR looks just about ready to go.  

- The [`trace-id` should be 16-bits per the spec](https://www.w3.org/TR/trace-context/#trace-id), so this change makes perfect sense. 
- Tests run with no failures both locally and [in the test suite](https://github.com/cloudfoundry/routing-release#-running-unit-and-integration-tests)

handlers/zipkin.go
n-shaw: @stijnvet  - It appears @seankellner's suggestion is more appropriate that grabbing `trace.Low` for the span.  Will you consider updating your PR with this change?  Thanks!

stijnvet: Yes, sorry it took so long to reply. Great suggestion

handlers/zipkin_test.go
geofffranks: @stijnvetstijnvet it looks like this is still in the zipkin_test.go but resolved in zipkin.go?

PR #305, opened 2022-01-12
Add tests to ensure that we support ALPN-enabled HTTP/1.1 clients

geofffranks:
In golang 1.17, [ALPN handling was altered ](https://github.com/golang/go/issues/46310) in a way that would
break HTTP/1.1 clients using HTTP, if the server only supported 'h2'
in its ALPN protocols. Previously, it would fall back to no-negotiated
mechanism. [Golang later added a short-circuit](https://github.com/golang/go/commit/dc00dc6c6bf3b5554e37f60799aec092276ff807#) specifically for ALPN
clients supporting http/1.1 with servers supporting h2 to fall-back as
if ALPN was not negotiated.

Modifying router.go to include "http/1.1" isn't strictly required
anymore, but is probably a better thing to do anyway, and it allows
us to validate what protocol was negotiated in our tests.

### DIFF
--- a/integration/main_test.go
+++ b/integration/main_test.go
@@ -347,12 +347,52 @@ var _ = Describe("Router Integration", func() {
 			Expect(err).ToNot(HaveOccurred())
 			Expect(resp.StatusCode).To(Equal(http.StatusOK))
 			Expect(resp.Proto).To(Equal("HTTP/2.0"))
+			Expect(resp.TLS.NegotiatedProtocol).To(Equal("h2"))
 
 			h1Client := &http.Client{Transport: &http.Transport{TLSClientConfig: clientTLSConfig}}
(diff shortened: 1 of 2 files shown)


integration/main_test.go
```diff
@@ -347,12 +347,53 @@ var _ = Describe("Router Integration", func() {
 			Expect(err).ToNot(HaveOccurred())
 			Expect(resp.StatusCode).To(Equal(http.StatusOK))
 			Expect(resp.Proto).To(Equal("HTTP/2.0"))
+			Expect(resp.TLS.NegotiatedProtocol).To(Equal("h2"))
 
+			// is this using ALPN? If so, why is it not failing with go 1.17???
```
ameowlia, line 352: Let's get rid of this comment

ameowlia (changes requested):
LGTM except for the comment

ameowlia (approved):

PR #307, opened 2022-01-21
Add a optional prometheus endpoint to record app latency

Benjamintf1:
Signed-off-by: Rebecca Roberts <robertsre@vmware.com>
Signed-off-by: Ben Fuller <benjaminf@vmware.com>
Signed-off-by: Andrew Crump <crumpan@vmware.com>

Note on necessary changes that will have to be made before making this a default feature:

* Possibly add more metadata, including app data, but also possibly success/failure outcome as well
* expire metrics after a period of time. 
*  There's a released change in loggregator-agent-release/metrics-discovery to handle bugs discovered to do with this change.

Add a optional Prometheus endpoint to record app latency

Prometheus latency should usually produce less metrics per application, but should also be easier to parse and resolve. It should also be more accurate. 

should be able to deploy with
```
  router:
       per_request_metrics_reporting: false
       send_http_start_stop_server_event: false
       send_http_start_stop_client_event: false
       per_app_prometheus_http_metrics_reporting: true
       prometheus:
         ca_cert: ((gorouter_metrics_tls.ca))
         cert: ((gorouter_metrics_tls.certificate))
         key: ((gorouter_metrics_tls.private_key))
         port: 18000
         server_name: gorouter_metrics
```
```
- name: gorouter_metrics_tls
  options:
    alternative_names:
    - gorouter_metrics
    ca: metric_scraper_ca
    common_name: gorouter_metrics
    extended_key_usage:
    - server_auth
  type: certificate
  update_mode: converge
```

prometheus metrics come out and the old latency metrics and start stops don't. 

prometheus metrics don't come out and old latency metrics and start stops do. 

### DIFF
--- a/handlers/httplatencyprometheus.go
+++ b/handlers/httplatencyprometheus.go
@@ -16,8 +16,7 @@ type httpLatencyPrometheusHandler struct {
 	registry Registry
 }
 
-// NewHTTPStartStop creates a new handler that handles emitting frontent
-// HTTP StartStop events
+// NewHTTPLatencyPrometheus creates a new handler that handles prometheus metrics for latency
 func NewHTTPLatencyPrometheus(r Registry) negroni.Handler {
 	return &httpLatencyPrometheusHandler{
 		registry: r,
--- a/handlers/httplatencyprometheus.go
+++ b/handlers/httplatencyprometheus.go
@@ -1,0 +1,47 @@
+package handlers
+
+import (
+	"net/http"
+	"time"
+
+	metrics "code.cloudfoundry.org/go-metric-registry"
+	"github.com/urfave/negroni"
+)
+
+type Registry interface {
+	NewHistogram(name, helpText string, buckets []float64, opts ...metrics.MetricOption) metrics.Histogram
+}
+
+type httpLatencyPrometheusHandler struct {
+	registry Registry
+}
+
+// NewHTTPStartStop creates a new handler that handles emitting frontent
+// HTTP StartStop events
+func NewHTTPLatencyPrometheus(r Registry) negroni.Handler {
+	return &httpLatencyPrometheusHandler{
+		registry: r,
+	}
+}
+
+// ServeHTTP handles emitting a StartStop event after the request has been completed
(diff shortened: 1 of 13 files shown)


handlers/httplatencyprometheus.go
```diff
@@ -0,0 +1,47 @@
...
+
+type Registry interface {
+	NewHistogram(name, helpText string, buckets []float64, opts ...metrics.MetricOption) metrics.Histogram
+}
+
+type httpLatencyPrometheusHandler struct {
+	registry Registry
+}
+
+// NewHTTPStartStop creates a new handler that handles emitting frontent
```
ctlong, line 19: I think this is the wrong name 😅

main.go
```diff
@@ -181,10 +181,17 @@ func main() {
 		logger.Fatal("new-route-services-server", zap.Error(err))
 	}
 
+	var metricsRegistry *mr.Registry
+	if c.Prometheus.Port != 0 {
+		metricsRegistry = mr.NewRegistry(log.Default(),
+			mr.WithTLSServer(int(c.Prometheus.Port), c.Prometheus.CertPath, c.Prometheus.KeyPath, c.Prometheus.CAPath))
```
geofffranks, line 187: We should probably have some validation that all four items are set, since they're required in conjunction with each other. Does it make more sense to have the if statement here depend on a feature-enabling flag? in other words, if we don't have `PerAppPrometheusHttpMetricsReporting` set, what is the point of setting up the registry or checking config values?

Benjamintf1: If PerAppPrometheusHttpMetricsReporting is set and prometheus isn't, then it fails during templating in the release. if you havn't set the certs, it'll fail when setting up the server. The cert path's themselves should be hardcoded if you have the port set in the release.

Benjamintf1: The point of setting up a registry without per_app_http prom metrics is that you still get the default metrics. Whether that be more metrics which we incorperate into the new method of emitting metrics, or the default metrics(questionable) that are emitted by go-metrics-registry.

handlers/httplatencyprometheus.go
Benjamintf1: fixed.

Benjamintf1:
Any more feedback you have? anything you want to do/want us to do before merging.

geofffranks (approved):

PR #316, opened 2022-05-09
Update Gorouter `CACerts` property to list

peanball:
This allows modifying and expanding the CAs in a development using ops-files.

**This is a breaking change in configuration and will require a new major version of CF Deployment!**

`CACerts` in the gorouter configuration is a string, containing concatenated PEM format CA certificates. In order to add new certificates, the default list of CA certificates provided in the `cf-deployment` should be _extended_ with additional certificates. Using `ops-files` in CF this is not possible.

By changing the `CACerts` field to an array, which may still contain PEM certificate chains, additional CA certificates can be added via ops-file without the risk of overwriting updated upstream CAs defined via `cf-deployment`.

See above.

The configuration allows defining CA Certificates as list of strings in the configuration.

The following ops-file to append an extra CA cert becomes possible:

```yaml
- type: replace
  path: /instance_groups/name=router/jobs/name=gorouter/properties/router/ca_certs?/-
  value:
     - ((YOUR-CUSTOM-CA-1.ca))
```

New CA Certs can be added via ops-file

New CA Certs can only be added by copy&pasting the default CA certs from the `cf-deployment` and adding a new one, risking to lose / overwrite changes introduced in a new `cf-deployment` manifest.

      * Some tests unrelated to this change fail in the current `main` branch.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -214,7 +214,7 @@ type Config struct {
 	EnableHTTP2     bool              `yaml:"enable_http2"`
 	SSLCertificates []tls.Certificate `yaml:"-"`
 	TLSPEM          []TLSPem          `yaml:"tls_pem,omitempty"`
-	CACerts         string            `yaml:"ca_certs,omitempty"`
+	CACerts         []string          `yaml:"ca_certs,omitempty"`
 	CAPool          *x509.CertPool    `yaml:"-"`
 	ClientCACerts   string            `yaml:"client_ca_certs,omitempty"`
 	ClientCAPool    *x509.CertPool    `yaml:"-"`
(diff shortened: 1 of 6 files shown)


ameowlia:
Hi @peanball, 

❓ Do you have a related PR for routing-release? Or else, I would think you would run into [this template check](https://github.com/cloudfoundry/routing-release/blob/b8c70780d43f9b751960c1df30c1022cf5fc47f1/jobs/gorouter/templates/gorouter.yml.erb#L360-L364).

In order to use this there needs to be changes in routing release to:
* gorouter spec property definition
* gorouter template
* gorouter template tests 

Thanks!

geofffranks:

peanball:
I will complete it next week. The PR is needed though. I would like to add a test for the new property.

On 31. May 2022, at 16:08, Amelia Downs ***@***.***> wrote:

﻿

Hi @peanball<https://github.com/peanball>,

❓ Do you have a related PR for routing-release? Or else, I would think you would run into this template check<https://github.com/cloudfoundry/routing-release/blob/b8c70780d43f9b751960c1df30c1022cf5fc47f1/jobs/gorouter/templates/gorouter.yml.erb#L360-L364>.

In order to use this there needs to be changes in routing release to:

  *   gorouter spec property definition
  *   gorouter template
  *   gorouter template tests

Thanks!

mariash (approved):
LGTM

mariash:
ran tests manually

PR #317, opened 2022-05-13
Improve Websocket Error Logging

a18e:
  * `x_cf_routererror` is now provided in access logs for failing websocket forwards

Easier analysis of websocket connection issues

  1. cause a websocket timeout, i.e. by deploying a websocket app that takes longer than 5s to upgrade
  2. observe logging

`access_log`:
 ```
js-ws.cf.local - [2022-05-13T13:06:57.725415361Z] "GET /slow HTTP/1.1" 502 0 0 "-" "-" "172.19.0.4:44124" "172.19.0.6:8080" x_forwarded_for:"172.19.0.1, 172.19.0.4" x_forwarded_proto:"https" vcap_request_id:"3688e056-611f-49c8-7d16-3b1b2b2fbc0d" response_time:12.585910 gorouter_time:12.585910 app_id:"-" app_index:"-" instance_id:"52d7d6f4-d32b-4091-5bd6-7cc5ba9142ac" x_cf_routererror:"endpoint_failure (timeout waiting for http response from backend)"
```

`access_log`:
 ```
js-ws.cf.local - [2022-05-13T13:10:48.064063585Z] "GET /slow HTTP/1.1" 502 0 0 "-" "-" "172.19.0.4:44842" "172.19.0.5:8080" x_forwarded_for:"172.19.0.1, 172.19.0.4" x_forwarded_proto:"https" vcap_request_id:"311a08de-cb3b-417c-7048-5f00f77c8d72" response_time:10.064082 gorouter_time:10.064082 app_id:"-" app_index:"-" instance_id:"acfab35f-73de-4f8f-4afd-66fd774b9be8" x_cf_routererror:"-"
```

https://github.com/cloudfoundry/gorouter/pull/319 - (was originally part of this PR)
https://github.com/cloudfoundry/gorouter/pull/294 - similar `x_cf_routererror` behaviour for a different `endpoint_failure`

  * Some tests unrelated to this change fail in the current `main` branch.

### DIFF
--- a/proxy/handler/forwarder.go
+++ b/proxy/handler/forwarder.go
@@ -25,7 +25,7 @@ type Forwarder struct {
 //
 // If the backend response code is not 101 Switching Protocols, then
 // ForwardIO will return immediately, allowing the caller to close the connections.
-func (f *Forwarder) ForwardIO(clientConn, backendConn io.ReadWriter) int {
+func (f *Forwarder) ForwardIO(clientConn, backendConn io.ReadWriter) (int, error) {
 	done := make(chan bool, 2)
 
 	copy := func(dst io.Writer, src io.Reader) {
@@ -45,7 +45,7 @@ func (f *Forwarder) ForwardIO(clientConn, backendConn io.ReadWriter) int {
 		if writeErr != nil {
 			f.Logger.Error("websocket-client-write", zap.Error(writeErr))
 		}
-		return http.StatusBadGateway
+		return http.StatusBadGateway, err
 	}
 
 	// as long as we got a valid response from the backend,
@@ -59,14 +59,14 @@ func (f *Forwarder) ForwardIO(clientConn, backendConn io.ReadWriter) int {
 		// but there isn't much we can do about that at this point
 		//
 		// return it so we can log it in access logs
-		return resp.StatusCode
+		return resp.StatusCode, err
 	}
 
 	if !isValidWebsocketResponse(resp) {
 		errMsg := fmt.Sprintf("backend responded with non-101 status code: %d", resp.StatusCode)
 		err = errors.New(errMsg)
 		f.Logger.Error("websocket-backend", zap.Error(err))
-		return resp.StatusCode
+		return resp.StatusCode, err
 	}
 
 	// only now do we start copying body data
@@ -75,7 +75,7 @@ func (f *Forwarder) ForwardIO(clientConn, backendConn io.ReadWriter) int {
 
 	// Note: this blocks until the entire websocket activity completes
 	<-done
-	return http.StatusSwitchingProtocols
+	return http.StatusSwitchingProtocols, nil
 }
 
 func isValidWebsocketResponse(resp *http.Response) bool {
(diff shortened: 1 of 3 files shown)


ameowlia:
Hi @a18e , 

Thank you so much for this PR! This PR includes two different bits of logic that I want to address separately. 

1. x_cf_routererror is now provided in access logs for failing websocket
This is a misuse of the x_cf_routererror header. This header is only supposed to be set when the gorouter itself errors and is there to indicate that there is something wrong with the platform. [Here is the list of current possible values for this header](https://docs.cloudfoundry.org/adminguide/troubleshooting-router-error-responses.html#gorouter-error-classification). "timeout waiting for http response from backend" seems most likely to be an app issue, not a platform issue.

❓ If it is difficult to debug websocket errors, maybe there is something else you could change? More logging perhaps? If you have an example app that produces an error that isn't exposed to the user, please share so we can look at the same example.

2. Add X-Vcap-Request-Id to stderr
  * Looks great!
  * Feel free to make this logic its own PR so we can merge it in sooner.


Thanks, Amelia

a18e:
Hi @ameowlia,

thanks for the review!

I've split off the `X-Vcap-Request-Id` change into a this PR: https://github.com/cloudfoundry/gorouter/pull/319.

Concerning your issues with the `x_cf_routererror` change:

There might be a misunderstanding here, perhaps I can clear it up a bit:

This PR does not change behaviour so that websocket app errors are being forwarded into `X-CF-RouterError`.

The only change is that if an error occurs inside GoRouter (`TimeoutError` in our case [1]) during the establishment of a websocket *connection* (I may have described this a bit unfortunately as "failing websocket **forwards**" because of the name of the class [2]), this error is being included in the access log.

I don't believe this is a misuse of `X-CF-RouterError` since the documentation states quite clearly that its specific purpose is to help differentiate between different "unknown" GoRouter errors which can be caused by either App or Platform (see last entry in the table you provided [3]).

From "Troubleshooting Router Error Responses - Gorouter specific Response Headers"[4] :

> In the case that Gorouter encounters an error connecting to an application backend, the `X-CF-RouterError` header will be populated to help distinguish the origin of a non-2xx response code.

There's also a specific mention of endpoint failures:

> | endpoint_failure | The registered endpoint for the desired route failed to handle the request. |
> | --- | --- |

We also set `X-CF-RouterError` to `endpoint_failure` if a "normal" (non-websocket) app does not respond in time.

To make it more clear that this is also an endpoint failure, I've modified the PR so that `X-CF-RouterError` is now set to `endpoint_failure (timeout waiting for http response from backend)` instead of just `timeout waiting for http response from backend` (for our example). This is similar behaviour to this merged PR: https://github.com/cloudfoundry/gorouter/pull/294. This way we also don't need to update the Troubleshooting Guide [2].

Let me know what you think.
Alex

[1] https://github.com/a18e/gorouter/blob/be23445bf155a4c4f8cfbc764eab6ff80f0c24bc/proxy/utils/response_reader.go#L18-L20

[2] https://github.com/cloudfoundry/gorouter/pull/317/files#diff-39c11590261633001d98421d0bee55047f9a3c66f62987e5d3b7b77e036d0e46

[3] https://docs.cloudfoundry.org/adminguide/troubleshooting-router-error-responses.html#gorouter-error-classification

[4] https://docs.cloudfoundry.org/adminguide/troubleshooting-router-error-responses.html#gorouter-specific-response-headers

ameowlia:
Thank you for this explanation @a18e and for the change in wording. I like the `endpoint_failure (timeout waiting for http response from backend)` better.



LGTM

PR #320, opened 2022-06-08
Retry in healthchecker instead of relying on monit

mariash:
## A short explanation of the proposed change:

Instead of involving monit poll cycles, retry in healthchecker before failing. Reconfigure monit to restart gorouter whenever healthchecker fails.

## An explanation of the use cases your change solves

When gorouter hangs, healthchecker should trigger the gorouter restart instead of just failing indefinitely.

## Instructions to functionally test the behavior change using operator interfaces (BOSH manifest, logs, curl, and metrics)

`kill -s SIGSTOP <GOROUTER_PID>`

## Expected result after the change

see that healthchecker fails within 30s and monit restarts gorouter immediately

## Current result before the change

gorouter is never restarted

## Links to any other associated PRs

https://github.com/cloudfoundry/routing-release/pull/275

* [x] I have made this pull request to the `develop` branch

* [x] I have run all the unit tests using `scripts/run-unit-tests-in-docker`

### DIFF
--- a/healthchecker/watchdog/watchdog.go
+++ b/healthchecker/watchdog/watchdog.go
@@ -10,9 +10,13 @@ import (
 	"time"
 
 	goRouterLogger "code.cloudfoundry.org/gorouter/logger"
+	"github.com/uber-go/zap"
 )
 
-const healthCheckEndpoint = "/healthz"
+const (
+	healthCheckEndpoint = "/healthz"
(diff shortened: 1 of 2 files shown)


jrussett:
Looks good to me. I tested this out and observed both the lack of restarting in the current release and the new restarting behavior with these changes.

PR #314, opened 2022-03-15
Return 503 instead of 404 for longer using pool timeout

kecirlotfi:
There is an open issue in cf/routing-release with id cloudfoundry/routing-release#250.
This proposal will change the way GoRouter manages routes for crashed apps.

We have three cases:
Below, for each case we describe, we will provide logs of tests before and after our change.
Before change is named `Current behaviour` and After change is named `New Behaviour`.

## The app is UP but not reachable by gorouter for some reason (ie if the endpoint uses TLS and the error is one of PrunableClassifiers),
The gorouter will remove the endpoint and once the pool is empty, it will be removed.
Then future calls on the app will get 404 response code.
Our change will prevent removing empty pool for `X` period of time. As result, all calls will respond with 503 code.
After that `X` period is passed, the empty pool will be removed. Gorouter will repond with 404 later on for this app.

This case was tested locally, after building gorouter

### Current behaviour

```
# register the route
nats-pub 'router.register' '{"host":"127.0.0.1","tls_port":1235,"uris":["tls.localhost.routing.cf-app.com"],"tags":{"another_key":"another_value","some_key":"some_value"}}'
```

```
#GoRouter logs
{"log_level":1,"timestamp":1648567660.59884,"message":"route-registered","source":"gorouter.stdout.registry","data":{"uri":"tls.localhost.routing.cf-app.com"}}
{"log_level":0,"timestamp":1648567660.598932,"message":"uri-added","source":"gorouter.stdout.registry","data":{"uri":"tls.localhost.routing.cf-app.com"}}
{"log_level":1,"timestamp":1648567660.5990021,"message":"endpoint-registered","source":"gorouter.stdout.registry","data":{"uri":"tls.localhost.routing.cf-app.com","backend":"127.0.0.1:1235","application_id":"","instance_id":"","server_cert_domain_san":"","protocol":"http1","modification_tag":{"guid":"","index":0},"isolation_segment":"-","isTLS":true}}
{"log_level":0,"timestamp":1648567664.4837272,"message":"vcap-request-id-header-set","source":"gorouter.stdout","data":{"VcapRequestIdHeader":"efb3b01d-aacf-4165-7de7-d2e53b66011b"}}
[...]

PR #321, opened 2022-07-09
Don't return empty pool if fallback exists

domdom82:

This PR fixes https://github.com/cloudfoundry/routing-release/issues/279

A regression was introduced with https://github.com/cloudfoundry/gorouter/pull/314 that returns 503 on routes with a path. If the path gets unmapped, gorouter returns 503 instead of falling back to the non-path route.

Mapping a more specific path `/foo` on an app is a valid use case for route services that may be used to rate limit that path only (instead of the whole app).

However, you still want to be able to use the less specific path `/` if you unmap the `/foo` route. Instead, gorouter gives you a 503 if `empty_pool_response_code_503 ` and  `empty_pool_timeout ` are enabled.

1. `cf push` any app. Make sure it responds 200 OK on `/foo`
2. `cf map-route my-app.cf-app.com --hostname my-app --path foo`
3. `cf unmap-route my-app.cf-app.com --hostname my-app --path foo`
4. `curl https://my-app.cf-app.com/foo`

```
200 OK (coming from my-app.cf-app.com)
```
```
503 Service Unavailable: Requested route ('my-app.cf-app.com') has no available endpoints.
```

https://github.com/cloudfoundry/gorouter/pull/314

### DIFF
--- a/registry/container/trie.go
+++ b/registry/container/trie.go
@@ -44,7 +44,7 @@ func (r *Trie) Find(uri route.Uri) *route.EndpointPool {
 	return nil
 }
 
-// MatchUri returns the longest route that matches the URI parameter, nil if nothing matches.
+// MatchUri returns the longest route that matches the URI parameter and has endpoints, nil if nothing matches.
 func (r *Trie) MatchUri(uri route.Uri) *route.EndpointPool {
 	key := strings.TrimPrefix(uri.String(), "/")
 	node := r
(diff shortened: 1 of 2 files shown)


geofffranks:

domdom82:
@jrussett any updates on this PR?

@geofffranks is there a way to re-trigger the CI?

domdom82:
@jrussett bump. do you need help reviewing this?

ameowlia (approved):

PR #322, opened 2022-07-13
fix: provide meaningful error message for rs error

maxmoehl:
* [X] I have run all the unit tests using `scripts/run-unit-tests-in-docker` from [routing-release](https://github.com/cloudfoundry/routing-release).

If a route-service violates the route-service timeout, the previous
error message was `400 Bad Request: Failed to validate Route Service
Signature for x-forwarded-client-cert` which is not indicative of the
underlying failure. This change does two things:
1. If the route-service validation failed because of a route-service
   timeout, respond with the proper error message and 504 GW timeout
2. If the route-service validation failed because of something else,
   respond with a 502 Bad GW since the error is most likely not on
   user side, but in the intermediate server.
I know that the status codes do not match 100%, but IMHO they are more
appropriate then the previous 400 Bad Request, and the route-service is,
in some sense, a upstream server to the gorouter, just not to the one
giving the error message.

Resolves: cloudfoundry/routing-release#272

### DIFF
--- a/handlers/clientcert.go
+++ b/handlers/clientcert.go
@@ -2,12 +2,15 @@ package handlers
 
 import (
 	"encoding/pem"
+	"errors"
+	"fmt"
 	"net/http"
 	"strings"
 
 	"code.cloudfoundry.org/gorouter/config"
 	"code.cloudfoundry.org/gorouter/errorwriter"
 	"code.cloudfoundry.org/gorouter/logger"
+	"code.cloudfoundry.org/gorouter/routeservice"
 	"github.com/uber-go/zap"
 	"github.com/urfave/negroni"
 )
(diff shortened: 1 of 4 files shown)


maxmoehl:
The new response for route-service timeouts would be:
```
504 Gateway Timeout: Failed to validate Route Service Signature: route service request expired
```

maxmoehl:
I ran the tests and got the test case related to my change green again, but there are still lots of other tests that were already failing.

geofffranks (commented):
Hey @maxmoehl!

I have a couple questions on this PR:

1. Does the misleading error message only occur in the case where a route-service is functional, but slow, resulting in a >60s delay between signing the message and the gorouter receiving  the signed request? Is the message when gorouter fails to contact the route-service at all handled properly?
2. Have you taken a look at the [code here](https://github.com/cloudfoundry/gorouter/blob/main/handlers/routeservice.go#L83-L92)? It also returns a similar status code and error message as part of the route-services handler. There might be another use case to be considered for this fix.
3. Are you able to provide tests for the case where a 504 is returned?
4. It looks like this gets rid of the test case for sanitizeError. How come?

maxmoehl:
Thanks for the review!

1. Will check that, but if I remember correctly it returns a 502 in that case
2. Didn't catch that, will update this section as well
3. Yes, that should be possible, will do
4. It doesn't, there are no cases for `sanitizeError`

I will try to prepare a commit for the open points today.

geofffranks:

@maxmoehl any progress on the updates for this?

maxmoehl:
Hi @geofffranks, sorry i totally forgot about this PR 😓 

I changed the second occurrence you pointed out and adapted the unit-tests, they are now passing locally (`./scripts/run-unit-tests-in-docker gorouter`). I also did a free-text search for more occurrences of such error handling but couldn't find any.

geofffranks (approved):

PR #325, opened 2022-08-26
Introduce separate setting for web socket upgrade dial timeout

peanball:

Websocket forwarding is essentially a new connection negotiation. So far the `endpoint_dial_timeout` value was used for websocket upgrades as well.

We have a case where the web socket upgrade takes longer than the default endpoint dial timeout because there is some processing happening before the websocket connection is accepted. Increasing the `endpoint_dial_timeout` can have negative repercussions for the regular use case, e.g. on backend failure retries.

Set the `websocket_dial_timeout` value to a specific value > 5s (default) and run with an endpoint that does a slow websocket upgrade, which responds in more than 5s (or the value for `endpoint_dial_timeout`) but faster than `websocket_dial_timeout`

The web socket request completes correctly.

The web socket request fails with a dial timeout, because the websocket dial timeout was not configurable separately.

https://github.com/cloudfoundry/routing-release/pull/284

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -246,6 +246,7 @@ type Config struct {
 	StartResponseDelayInterval      time.Duration `yaml:"start_response_delay_interval,omitempty"`
 	EndpointTimeout                 time.Duration `yaml:"endpoint_timeout,omitempty"`
 	EndpointDialTimeout             time.Duration `yaml:"endpoint_dial_timeout,omitempty"`
+	WebsocketDialTimeout            time.Duration `yaml:"websocket_dial_timeout,omitempty"`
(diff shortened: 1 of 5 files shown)


geofffranks:
@plowin That is accurate. I updated repo settings, can you see if you're able to merge this now?

plowin:
Hi @geofffranks , I got the permissions now and could merge.
If the CI can be fixed in the near future, we could wait for this as I did not cross-ran the tests locally yet.

geofffranks:
i don't think ci will be fixed soon. i'm contemplating rolling back the change completely and starting over 😬

PR #326, opened 2022-08-30
Fix default config initialization for websocket dial timeout value

peanball:
The value is either defined explicitly in the configuration, or derived from the `endpoint_dial_timeout`.
Because the config read from file is applied on top of the values in `Config.defaultConfig`, the
`WebsocketDialTimeout` must not be defined there yet, as it would disable the automatic derivation of the value from a configured `endpoint_dial_timeout` and would change behavor.

The gorouter tests use the `defaultConfig` without `Process()`, so the value was never derived and `WebsocketDialTimeout` remained at `0`, failing some tests.

Where needed, the value is now set explicitly in the test harness, but for regular operation it is derived if not set, as described above.

Also fixes long-standing flaky tests `Router Integration route services when the route service is hosted on the platform (internal, has a route as an app)`, which seems to have been a timing issue.

Fix for #325, which broke some tests due to missing config value propagation for a new property. `websocket_dial_timeout` falls back to the configured value of `endpoint_dial_timeout` when `websocket_dial_timeout` is not configured. The logic for setting the value when it was not configured was not called in tests.

### DIFF
--- a/integration/main_test.go
+++ b/integration/main_test.go
@@ -805,9 +805,11 @@ var _ = Describe("Router Integration", func() {
 				req, err := http.NewRequest("GET", fmt.Sprintf("http://%s:%d", localIP, proxyPort), nil)
 				Expect(err).ToNot(HaveOccurred())
 				req.Host = "demo." + test_util.LocalhostDNS
-				res, err := client.Do(req)
-				Expect(err).ToNot(HaveOccurred())
-				Expect(res.StatusCode).To(Equal(TEST_STATUS_CODE))
+				Eventually(func() int {
+					res, err := client.Do(req)
+					Expect(err).ToNot(HaveOccurred())
+					return res.StatusCode
+				}, 5*time.Second).Should(Equal(TEST_STATUS_CODE))
 			})
 
 			Context("when the client connects with HTTPS", func() {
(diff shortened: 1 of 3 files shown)


integration/main_test.go
```diff
@@ -805,9 +805,11 @@ var _ = Describe("Router Integration", func() {
 				req, err := http.NewRequest("GET", fmt.Sprintf("http://%s:%d", localIP, proxyPort), nil)
 				Expect(err).ToNot(HaveOccurred())
 				req.Host = "demo." + test_util.LocalhostDNS
-				res, err := client.Do(req)
-				Expect(err).ToNot(HaveOccurred())
-				Expect(res.StatusCode).To(Equal(TEST_STATUS_CODE))
+				Eventually(func() int {
```
plowin, line 808: For me, the integration-suite tests have different issues locally before and after this change.
Maybe it depends on the tester-system, keeping it in this PR as it can only improve the current status from my POV

plowin (approved):
thanks a lot for the fast fixup!

integration/main_test.go
peanball: Before I had consistent failures on this particular test, now it's rare.

plowin: True, same for me, thx for this enhancement!

PR #331, opened 2022-10-10
Emit access logs for 431 responses to Loggegator

dsabeti:
**Background**
Even though access logs for 431 responses were correctly being logged to `access.log`, these logs were not being emitted to Loggregator. The code for emitting these logs [requires an AppID or a SourceInstanceID](https://github.com/cloudfoundry/gorouter/blob/379860daa83a162ffe0b6039eafb7c8bfa1eaccf/accesslog/dropsonde_logsender.go#L24-L26). However, because the handler that enforces request header size returns a 431 before an app/backend is even selected, these fields were both missing, and the log was not being emitted.

**The Change**
This change adds logic to the MaxRequestSize handler so that it chooses a backend in order to populate the AppID field and successfully emits the access log to Loggregator. This backend is only chosen in cases where it will cut the request short and return a 431.

*Other smaller/incidental changes*
- Logs a message to gorouter log files when a Loggregator log message is not sent due to AppID or SourceInstanceID being empty
- Adding a new helper function `EndpointIteratorForRequest` for reuse in multiple files/packages
- Moving some code from the `proxy` package to the `handlers` package so that the new helper function can leverage the code without a circular import
- Some reorganization of tests in `max_request_size_test.go`
- Additional functionality for the `fakeMetron` in integration tests for easier testing

### DIFF
--- a/accesslog/dropsonde_logsender.go
+++ b/accesslog/dropsonde_logsender.go
@@ -1,6 +1,7 @@
 package accesslog
 
 import (
+	"fmt"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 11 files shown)


ebroberson (approved):
LGTM

PR #333, opened 2022-11-07
Fix memory leak when connections are prematurely closed on the client…

geofffranks:
…-side.

If a client closes a connection while gorouter is waiting on a response from the app, we are leaking request objects. When the request objects are up to 1mb each, it becomes easy to grow gorouter's memory footprint, requiring a restart.

This ensures that in the event of the 499 cases, we close the original request body, allowing the full request to be garbage collected.


reneighbor:
+1

PR #340, opened 2023-03-15
Add trace identifier to gorouter logs when handling request

mariash:
**A short explanation of the proposed change:**

Every request handler in the Gorouter now adds unique trace and span ID to the logs. 

Trace and span IDs are consolidated from B3, W3C and VCAP_REQUEST_ID headers.

Details on how consolidation of headers works:

- If B3 headers are set by user and zipkin is enabled in Gorouter, if they can be parsed as valid UUID, they will be used for trace and span ID as well as for VCAP_REQUEST_ID and W3C ID.
- If B3 headers are not set and zipkin is enabled, Gorouter will generate trace and span ID. They will be used for B3, VCAP_REQUEST_ID and W3C ID.
- If zipkin is disabled, and W3C headers are set by user and W3C is enabled in the Gorouter, W3C headers will be used as trace and span ID as well as for VCAP_REQUEST_ID.
- If W3C headers are not set by user and W3C is enabled in the Gorouter, W3C headers will be generated in the Gorouter and set as trace and span ID as well as for VCAP_REQUEST_ID.
- VCAP_REQUEST_ID cannot be set by user. It will be generated and used as trace and span ID if zipkin and W3C are disabled in the gorouter.

**An explanation of the use cases your change solves**

It would be easier to trace a specific request across multiple logs - gorouter.log, gorouter.err.log, access.log. In the future, the same trace ID will be used for the logs that follows a transaction in the system across multiple components (Cloud Controller, BBS, Rep, Garden) like cf push. This trace ID can be used to search all related logs to the transaction in log accumulating systems like splunk.

**Instructions to functionally test the behavior change using operator interfaces (BOSH manifest, logs, curl, and metrics)**

```
$ curl api.cf-app.com
```

On the gorouter VM, tail all logs in `/var/vcap/sys/log/gorouter`. See logs contain:

```
"data":{"trace-id":"25f23d6af46d460e71357ddc0759a198","span-id":"71357ddc0759a198"}
```
**Expected result after the change**

All gorouter logs related to request handling have trace and span ID.

**Current result before the change**

Only access.log contain request ID.

* [X] I have run all the unit tests using `scripts/run-unit-tests-in-docker` from [routing-release](https://github.com/cloudfoundry/routing-release).

### DIFF
--- a/handlers/access_log.go
+++ b/handlers/access_log.go
@@ -55,7 +55,7 @@ func (a *accessLog) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http
 
 	reqInfo, err := ContextRequestInfo(r)
 	if err != nil {
-		a.logger.Fatal("request-info-err", zap.Error(err))
+		a.logger.Panic("request-info-err", zap.Error(err))
 		return
 	}
 
--- a/handlers/access_log_test.go
+++ b/handlers/access_log_test.go
@@ -150,9 +150,9 @@ var _ = Describe("AccessLog", func() {
 			handler.Use(handlers.NewAccessLog(accessLogger, extraHeadersToLog, false, fakeLogger))
 			handler.Use(nextHandler)
 		})
-		It("calls Fatal on the logger", func() {
+		It("calls Panic on the logger", func() {
 			handler.ServeHTTP(resp, req)
-			Expect(fakeLogger.FatalCallCount()).To(Equal(1))
+			Expect(fakeLogger.PanicCallCount()).To(Equal(1))
 		})
 	})
 
--- a/handlers/clientcert.go
+++ b/handlers/clientcert.go
@@ -42,6 +42,7 @@ func NewClientCert(
 }
 
 func (c *clientCert) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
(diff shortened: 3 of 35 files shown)


domdom82:
@mariash how does this behave across route service calls? e.g. if gorouter get a request for a route that has a route service attached, it will forward the request to the route service first. the request usually ends up again at gorouter, now with `host` set to the host of the route service. When the route service decides to forward the request to the backend app, there will be another request to yet another gorouter, this time targeting the backend app.

Currently, there will be three different VCAP_REQUEST_IDs used for each of the requests and it is relatively hard to correlate them. Will this PR keep a single trace-id across all of them instead? Reading your description, they should be set by the user (in this case the use is gorouter itself), so they would replace VCAP_REQUEST_ID on subsequent requests?

ebroberson (approved):
LGTM

reneighbor:
LGTM too

mariash:
@domdom82 It will be the same VCAP_REQUEST_ID in the following case:
1. Routers should have [zipkin](https://github.com/cloudfoundry/routing-release/blob/develop/jobs/gorouter/spec#L292) or [w3c](https://github.com/cloudfoundry/routing-release/blob/develop/jobs/gorouter/spec#L295) enabled.
2. Route services pass down b3 or w3c trace id to the gorouter.

PR #342, opened 2023-04-13
fix(trace): remove GotConn delay

maxmoehl:
The delay causes a race condition in the go transport that results in a 502 Bad Gateway with:
  `endpoint_failure (readLoopPeekFailLocked: %!w(<nil>))`.

This happens because the transport peeks the first few bytes on the connection and gets some data even though it doesn't expect any. This causes it to go into an error state even though there is no error resulting in the formatting directive to break.

This commit removes the delay and adds a note why we can't do this for now. This will reduce the amount of requests we can retry because the client will send data before we know that the connection is good. After we sent _some_ data we can't be sure that the server hasn't started processing, hence no retry in such cases.

See: https://cloudfoundry.slack.com/archives/C033ALST37V/p1680888356483179
See: golang/go#31259
Resolves: cloudfoundry/routing-release#316

### DIFF
--- a/proxy/round_tripper/trace.go
+++ b/proxy/round_tripper/trace.go
@@ -92,12 +92,14 @@ func traceRequest(req *http.Request) (*http.Request, *requestTracer) {
 		GotConn: func(info httptrace.GotConnInfo) {
 			t.gotConn.Store(true)
 			t.connInfo.Store(&info)
-			if !info.Reused {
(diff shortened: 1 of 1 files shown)


domdom82:
technically, the race condition is already there. the delay only makes sure it is predictable by allowing one go routine to be scheduled before the competing one. so the delay does *cause* the race. Also bear in mind that even without the delay problems are not solved and the issue raised by @Gerg may still appear. it's just more random now.

PR #337, opened 2023-03-02
feat(round-tripper): retry requests more often

maxmoehl:
The round tripper didn't make any effort to understand at which stage a
request failed and consequently only retried requests that failed while
establishing the connection or are idempotent by definition. This
commit introduces a check whether the headers were fully transmitted to
the target endpoint. If the headers were not transmitted, the target
server should discard the request in any case allowing us to retry it.

#### Check-List:
* [x] I have viewed signed and have submitted the Contributor License Agreement
* [x] I have made this pull request to the `main` branch
* [x] I have run all the unit tests using `scripts/run-unit-tests-in-docker` from [routing-release](https://github.com/cloudfoundry/routing-release).
* [ ] (Optional) I have run Routing Acceptance Tests and Routing Smoke Tests on bosh lite
* [ ] (Optional) I have run CF Acceptance Tests on bosh lite

PR #345, opened 2023-05-08
fix: do not retry if context has been cancelled

maxmoehl:
The additional retry logic introduced by #337 caused any error to become retry-able if the request did not make it to the back end. This included errors that came from context cancellation. This commit makes requests non-retry-able as soon as the context has been cancelled.

See related Slack discussion: https://cloudfoundry.slack.com/archives/C01ABMVNE9E/p1683313910835329

### DIFF
--- a/proxy/round_tripper/proxy_round_tripper.go
+++ b/proxy/round_tripper/proxy_round_tripper.go
@@ -441,6 +441,11 @@ func isIdempotent(request *http.Request) bool {
 }
 
 func (rt *roundTripper) isRetriable(request *http.Request, err error, trace *requestTracer) (bool, error) {
+	// if the context has been cancelled we do not perform further retries
+	if request.Context().Err() != nil {
+		return false, fmt.Errorf("%w (%w)", request.Context().Err(), err)
+	}
+
 	// io.EOF errors are considered safe to retry for certain requests
 	// Replace the error here to track this state when classifying later.
 	if err == io.EOF && isIdempotent(request) {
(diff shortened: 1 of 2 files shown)


maxmoehl:
This fix should address the issues reported in the linked Slack thread. Nevertheless we should make the [classifier lists](https://github.com/cloudfoundry/gorouter/blob/main/proxy/fails/classifier_group.go) distinct and re-consider under which circumstances an endpoint should be removed from the pool, but this is out of scope for this PR.

maxmoehl:
Running the full unit test suite in docker fails because the image is gone:
```
Error response from daemon: pull access denied for cloudfoundry/cf-routing-pipeline, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -441,6 +441,11 @@ func isIdempotent(request *http.Request) bool {
 }
 
 func (rt *roundTripper) isRetriable(request *http.Request, err error, trace *requestTracer) (bool, error) {
+	// if the context has been cancelled we do not perform further retries
+	if request.Context().Err() != nil {
+		return false, request.Context().Err()
```
maxmoehl, line 446: As an alternative we could also return the error that has been passed in or combine the two errors to not loose any information:

```suggestion
		return false, fmt.Errorf("%w (%w)", err, request.Context().Err())
```

I'm open for suggestions.

geofffranks: i like this suggestion. let's wrap the error to provide all the info.

maxmoehl: I changed it, but swapped the errors since `context cancelled` is the main reason we are stopping the processing of the request.

PR #352, opened 2023-07-03
add test for 100-continue access logs

ameowlia:
In this PR -- https://github.com/cloudfoundry/gorouter/pull/351 -- I updated the `proxy/utils` so that 1XX status codes would not be written to the access logs when followed correctly by another non-1XX status code. That was tested at the unit level. This PR adds a higher level test for this same functionality.


reneighbor (approved):
LGTM. Nice test!

PR #353, opened 2023-07-12
feat(gorouter): retry requests on 529 status code.

domdom82:

Some users want to fine-grain the flow of requests to their apps.
Concrete example: An app can take one request, but is then busy for a while so it wants the client to go to another instance.
Usually, you would send a `429 Too many requests` and let the client do a retry.
However, in this case the client can't retry on a different instance, because Gorouter selects the instance.

This change makes it transparent for the client in that a backend may opt to send a `529 Service is overloaded` as a hint for gorouter to try another instance. The client does not need special coding, unless *all* of the backends are overloaded (in which case the 529 will be forwarded to the client). By default, gorouter will transparently retry on 529s as if any other retry-able error had occurred.

See above. App users want to be able to tell gorouter deliberately to retry another instance.

tbd

Gorouter retries on 529s

Gorouter ignores 592s

n/a

### DIFF
--- a/proxy/fails/basic_classifiers.go
+++ b/proxy/fails/basic_classifiers.go
@@ -13,6 +13,8 @@ var IdempotentRequestEOFError = errors.New("EOF (via idempotent request)")
 
 var IncompleteRequestError = errors.New("incomplete request")
 
+var BackendOverloadedError = errors.New("backend overloaded")
+
 var AttemptedTLSWithNonTLSBackend = ClassifierFunc(func(err error) bool {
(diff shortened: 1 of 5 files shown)


domdom82:
Closing this draft as it was deemed too risky. The largest risk was identified as the need to pre-buffer the request body. It is unclear how large that buffer should be as it depends on the nature of traffic, there is no "one size fits all" solution here. The second issue is that even if a modest limit was chosen (e.g. 1 mb), there is no stopping a client from sending hundreds of large requests that would all have to be buffered on gorouter, opening the door for a potential DOS attack.

PR #356, opened 2023-08-04
Continue to forward relevant headers

winkingturtle-vmw:
by removing hop-by-hop headers that could interfere with gorouter.

Context:
    - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Connection
    - https://nathandavison.com/blog/abusing-http-hop-by-hop-request-headers

The following list of keywords are removed from Connection header, so that gorouter can continue to utilize those headers when they are set and will continue to forward them down to the relavant handlers.

nits: Use io.ReadAll instead of ioutil since that's been the default as of Go1.16

### DIFF
--- a/proxy/handler/request_handler.go
+++ b/proxy/handler/request_handler.go
@@ -165,6 +165,40 @@ func (h *RequestHandler) HandleWebSocketRequest(iter route.EndpointIterator) {
 	h.reporter.CaptureWebSocketUpdate()
 }
 
+func (h *RequestHandler) SanitizeRequestConnection() {
+	connections := h.request.Header.Values("Connection")
+	for index, connection := range connections {
+		if connection != "" {
+			values := strings.Split(connection, ",")
+			connectionHeader := []string{}
+			bannedList := []string{
+				"X-Forwarded-For",
+				"X-Forwarded-Proto",
+				"B3",
+				"X-B3",
+				"X-B3-SpanID",
+				"X-B3-TraceID",
+				"X-Request-Start",
+				"X-Forwarded-Client-Cert",
+			}
+			for i := range values {
+				trimmedValue := strings.TrimSpace(values[i])
+				found := false
+				for _, item := range bannedList {
+					if strings.ToLower(item) == strings.ToLower(trimmedValue) {
+						found = true
+						break
+					}
+				}
+				if !found {
+					connectionHeader = append(connectionHeader, trimmedValue)
+				}
+			}
+			h.request.Header[http.CanonicalHeaderKey("Connection")][index] = strings.Join(connectionHeader, ", ")
+		}
+	}
+}
+
 type connSuccessCB func(net.Conn, *route.Endpoint) error
 type connFailureCB func(error)
 
--- a/proxy/handler/request_handler.go
+++ b/proxy/handler/request_handler.go
@@ -169,6 +169,9 @@ func (h *RequestHandler) HandleWebSocketRequest(iter route.EndpointIterator) {
 }
 
 func (h *RequestHandler) SanitizeRequestConnection() {
+	if len(h.hopByHopHeadersToFilter) == 0 {
+		return
+	}
 	connections := h.request.Header.Values("Connection")
 	for index, connection := range connections {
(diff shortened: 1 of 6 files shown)


proxy/handler/request_handler.go
```diff
@@ -165,6 +165,40 @@ func (h *RequestHandler) HandleWebSocketRequest(iter route.EndpointIterator) {
 	h.reporter.CaptureWebSocketUpdate()
 }
 
+func (h *RequestHandler) SanitizeRequestConnection() {
+	connections := h.request.Header.Values("Connection")
+	for index, connection := range connections {
+		if connection != "" {
+			values := strings.Split(connection, ",")
+			connectionHeader := []string{}
+			bannedList := []string{
```
domdom82, line 174: hi @winkingturtle-vmw thanks for your PR!
1. We want to keep these headers and protect them from getting dropped as a hop-by-hop, right? I think it may be a bit confusing to call the list `bannedList` then. Maybe `keepHeaders` or `headerToKeep`?

2. I think it makes sense to not have this list hard-coded as it may be different depending on the organization where CF is deployed.
I suggest to make it configurable in the [config](https://github.com/cloudfoundry/gorouter/blob/main/config/config.go) and provide a [spec](https://github.com/cloudfoundry/routing-release/blob/develop/jobs/gorouter/spec) for it.

You may provide the hard-coded list here as a default value for the spec and also put it into the [default config](https://github.com/cloudfoundry/gorouter/blob/main/config/config.go#L318) so that it is active by default.

domdom82 (changes requested):

proxy/handler/request_handler.go
winkingturtle-vmw: Thanks @domdom82 for you feedback. I think the spec property is an option, but can I please ask you with the task coming up with a good name? 

I am happy to change the name from `bannedList` to anything folks think is more easily understood, but essentially those header values would be banned from Client side on `Connection` header and are subject to removal. I know for me personally, `headersToKeep` would make it sound like a property that would need to have every single header used in gorouter.

domdom82: @winkingturtle-vmw what about `connectionHeadersToKeep`? I think it fits well with `connectionHeader` on line 173 and expresses that these are headers on the connection we want to keep instead of drop.

winkingturtle-vmw: I think between the two options, I would prefer `headersToKeep` probably because the values in this field are to be removed from `Connection` header instead of kept. Alternatively, I could just call this variable in this function a `list` to avoid any confusion. That said, in order to make this a property of gorouter, we would still need to give it a name that I am not sure what to call.

domdom82: ok, then I'll make the call to name it `headersToKeep` for now. In the spec you can add more description that explains in detail what the thing does.

winkingturtle-vmw: The variable was no longer needed after adding the spec property. It's now called `HopByHopHeadersToFilter` property. Is that okay if we now resolve this conversation?

proxy/handler/request_handler.go
```diff
@@ -165,6 +168,30 @@ func (h *RequestHandler) HandleWebSocketRequest(iter route.EndpointIterator) {
 	h.reporter.CaptureWebSocketUpdate()
 }
 
+func (h *RequestHandler) SanitizeRequestConnection() {
```
domdom82, line 171: this 3x nested for-loop will run on every request, regardless of the `HopByHopHeadersToFilter` config setting atm.
can we make sure it runs only when there are actual values in the config property? don't want to waste time on a critical path.

something like
```suggestion
func (h *RequestHandler) SanitizeRequestConnection() {
if len(h.hopByHopHeadersToFilter) == 0 {
  return
}
```

domdom82 (changes requested):

jrussett (approved):
These changes look good to me. I was able to deploy this code (via https://github.com/cloudfoundry/routing-release/pull/331) and everything worked correctly there. Merging this in.

PR #363, opened 2023-09-07
fix(proxy): Add missing timings to non-http backend requests

domdom82:

Fixes [gorouter_time is equal to response_time](https://github.com/cloudfoundry/routing-release/issues/278)

The request_handler responsible for websocket backends does not implement timing events, yet. This leads to gorouter_time being equal to response_time as there is no app_time recorded.

The same is true for all attempt details (backend_time, dial_time, tls_time, dns_time and failed_attempts_time) if `enable_log_attempts_details` is enabled.

- Correct timings also for non-HTTP backends. `gorouter_time` should only include time spent within the gorouter process, not time spent at (websocket-) backends.
- Measure attempt details for websocket requests, too. This includes `backend_time`, `dial_time`, `tls_time`, `failed_attempts_time` and `failed_attempts` and should behave like the measurements taken for regular HTTP requests in ProxyRoundTripper. `dns_time` will not be recorded because dns lookups will only ever occur on route service URLs.

1. push a websocket app
2. curl it
3. (optional) turn `enable_log_attempts_details` on 

`gorouter_time` and `response_time` should differ (with `response_time` being much larger than `gorouter_time`)

(optional)

`backend_time`, `dial_time` and `tls_time` should be present in the access log

`gorouter_time` and `response_time` are equal

(optional)

`backend_time`, `dial_time` and `tls_time` are not present in the access log

[Companion PR to vendor golang.org/x/net/websocket](https://github.com/cloudfoundry/routing-release/pull/354) (needs to be merged first)

### DIFF
--- a/proxy/proxy_test.go
+++ b/proxy/proxy_test.go
@@ -1346,9 +1346,10 @@ var _ = Describe("Proxy", func() {
 			defer ln.Close()
 
 			conn := dialProxy(proxyServer)
-			wsConn := wsClient(conn, "ws://ws-test")
+			wsConn, err := wsClient(conn, "ws://ws-test")
+			Expect(err).NotTo(HaveOccurred())
 
-			_, err := wsConn.Write([]byte("HELLO WEBSOCKET"))
+			_, err = wsConn.Write([]byte("HELLO WEBSOCKET"))
 			Expect(err).NotTo(HaveOccurred())
 
 			msgBuf := make([]byte, 100)
@@ -1387,13 +1388,13 @@ var _ = Describe("Proxy", func() {
 					_, err := http.ReadRequest(conn.Reader)
 					Expect(err).NotTo(HaveOccurred())
 
-					body := "this body took two seconds."
+					body := "this body took some time."
 					conn.WriteLines([]string{
 						"HTTP/1.1 200 OK",
 						fmt.Sprintf("Content-Length: %d", len(body)),
 					})
 
-					time.Sleep(2 * time.Second)
+					time.Sleep(100 * time.Millisecond)
 
 					conn.WriteLine(body)
 					conn.Close()
@@ -1426,8 +1427,8 @@ var _ = Describe("Proxy", func() {
 				Expect(strings.HasPrefix(string(b), "slow-response-app - [")).To(BeTrue())
 				Expect(string(b)).To(ContainSubstring("response_time:"))
 				Expect(string(b)).To(ContainSubstring("gorouter_time:"))
-				Expect(string(b)).To(MatchRegexp("response_time:2"))
-				Expect(string(b)).To(MatchRegexp("gorouter_time:0"))
+				Expect(string(b)).To(MatchRegexp("response_time:0.1"))
+				Expect(string(b)).To(MatchRegexp("gorouter_time:0.00"))
 			})
 
 			Context("in websocket requests", func() {
@@ -1438,7 +1439,7 @@ var _ = Describe("Proxy", func() {
 						Expect(err).NotTo(HaveOccurred())
 						Expect(string(msgBuf[:n])).To(Equal("HELLO WEBSOCKET"))
 
-						time.Sleep(time.Second)
+						time.Sleep(100 * time.Millisecond)
 
 						_, _ = conn.Write([]byte("WEBSOCKET OK"))
 						conn.Close()
@@ -1446,9 +1447,10 @@ var _ = Describe("Proxy", func() {
 					defer ln.Close()
 
 					conn := dialProxy(proxyServer)
-					wsConn := wsClient(conn, "ws://slow-ws-test")
+					wsConn, err := wsClient(conn, "ws://slow-ws-test")
+					Expect(err).NotTo(HaveOccurred())
 
-					_, err := wsConn.Write([]byte("HELLO WEBSOCKET"))
+					_, err = wsConn.Write([]byte("HELLO WEBSOCKET"))
 					Expect(err).NotTo(HaveOccurred())
 
 					msgBuf := make([]byte, 100)
@@ -1473,14 +1475,14 @@ var _ = Describe("Proxy", func() {
 					Expect(strings.HasPrefix(logStr, "slow-ws-test - [")).To(BeTrue())
 					Expect(logStr).To(ContainSubstring(`"GET / HTTP/1.1" 101`))
 					Expect(logStr).To(ContainSubstring(`x_forwarded_for:"127.0.0.1" x_forwarded_proto:"http" vcap_request_id:`))
-					Expect(logStr).To(MatchRegexp(`response_time:1`))
-					Expect(logStr).To(MatchRegexp(`gorouter_time:0`))
+					Expect(logStr).To(MatchRegexp(`response_time:0.1`))
+					Expect(logStr).To(MatchRegexp(`gorouter_time:0.00`))
 				})
 			})
 
 			Context("A slow app with multiple broken endpoints and attempt details logging enabled", func() {
 				BeforeEach(func() {
-					conf.EndpointDialTimeout = 1 * time.Second
+					conf.EndpointDialTimeout = 100 * time.Millisecond
 					conf.Logging.EnableAttemptsDetails = true
 					conf.DropletStaleThreshold = 1
 				})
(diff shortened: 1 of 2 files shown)


domdom82:
Note to self:
I just realized the entire websocket handling is broken with Envoy in place:
- The [for-loop](https://github.com/cloudfoundry/gorouter/blob/main/proxy/handler/request_handler.go#L225) in request_handler only takes tcp and tls dialling into account, not http
- The retry-logic only works on the tcp / tls level which almost always succeeds because Envoy will always accept a connection, even if the backend WebSocket app is down
- The http request (the one sending the upgrade header) only happens [in a callback](https://github.com/cloudfoundry/gorouter/blob/main/proxy/handler/request_handler.go#L150) after the retry-loop has exited
- With Envoy in place and a down backend, this will immediately result in a io.EOF error that will never be retried

So, to get a similar behavior like we have with proxy_round_tripper we will need to put at least tcp,tls + http in the retry-loop:
- [dial](https://github.com/cloudfoundry/gorouter/blob/main/proxy/handler/request_handler.go#L237)
- [send request](https://github.com/cloudfoundry/gorouter/blob/main/proxy/handler/request_handler.go#L262)

the [io copy](https://github.com/cloudfoundry/gorouter/blob/main/proxy/handler/request_handler.go#L275) is technically no longer HTTP, so we may ignore errors occurring there with respect to retry-ability of the request (I'm open for debate on this, though) 

this should resemble a `roundTrip` call that we also guard with retries in regular HTTP requests.

proxy/proxy_test.go
```diff
@@ -1385,6 +1431,55 @@ var _ = Describe("Proxy", func() {
...
 
+			Context("in websocket requests", func() {
+				It("shows slowness in response_time, not gorouter_time", func() {
+					ln := test_util.RegisterWSHandler(r, "slow-ws-test", func(conn *websocket.Conn) {
+						msgBuf := make([]byte, 100)
+						n, err := conn.Read(msgBuf)
+						Expect(err).NotTo(HaveOccurred())
+						Expect(string(msgBuf[:n])).To(Equal("HELLO WEBSOCKET"))
+
+						time.Sleep(time.Second)
```
ameowlia, line 1442: Can you re-write this test to not require a sleep? We record the gorouter time with 6 decimal spots. I think it is accurate enough that gorouter_time and response_time will be different, even without a sleep. If there _must_ be a sleep due to flakiness (maybe your machine is very fast!) then can we have millisecond sleeps?

proxy/proxy_test.go
```diff
@@ -1385,6 +1431,55 @@ var _ = Describe("Proxy", func() {
...
+					//make sure the record includes all the data
+					//since the building of the log record happens throughout the life of the request
+					logBytes, err := os.ReadFile(f.Name())
+					Expect(err).NotTo(HaveOccurred())
+					logStr := string(logBytes)
+					Expect(strings.HasPrefix(logStr, "slow-ws-test - [")).To(BeTrue())
+					Expect(logStr).To(ContainSubstring(`"GET / HTTP/1.1" 101`))
+					Expect(logStr).To(ContainSubstring(`x_forwarded_for:"127.0.0.1" x_forwarded_proto:"http" vcap_request_id:`))
+					Expect(logStr).To(MatchRegexp(`response_time:1`))
+					Expect(logStr).To(MatchRegexp(`gorouter_time:0`))
```
ameowlia, line 1479: Perhaps instead of checking for `1` and `0`, you can get the values and make sure that the response_time > gorouter_time.

proxy/proxy_test.go
```diff
@@ -1451,6 +1546,145 @@ var _ = Describe("Proxy", func() {
...
+						test_util.RegisterAddr(r, "partially-broken-ws-app", "10.255.255.1:1234", test_util.RegisterConfig{InstanceIndex: "1"})
+						test_util.RegisterAddr(r, "partially-broken-ws-app", "10.255.255.1:1235", test_util.RegisterConfig{InstanceIndex: "2"})
+
+						ln := test_util.RegisterWSHandler(r, "partially-broken-ws-app", func(conn *websocket.Conn) {
+							msgBuf := make([]byte, 100)
+							n, err := conn.Read(msgBuf)
+							Expect(err).NotTo(HaveOccurred())
+							Expect(string(msgBuf[:n])).To(Equal("HELLO WEBSOCKET"))
+
+							time.Sleep(2 * time.Second)
```
ameowlia, line 1607: Similar to above, I'd like to lessen this sleep. I see why you did it because you are testing these different times. Any chance you can lower it to some fraction of a second instead of 2 seconds?

ameowlia (changes requested):
1. ✅ I validated the issue you were seeing using your app.

Here is an example of a websocket connection where the `response_time` and the `gorouter_time` are equal.
```
2023-10-04T21:11:15.43+0000 [RTR/0] OUT go-responder.band-160711.cf-app.com - [2023-10-04T21:10:54.808352375Z] "GET /ws HTTP/1.1" 101 0 0 "-" "curl/7.81.0" "10.0.1.9:21418" "10.0.1.14:61002" x_forwarded_for:"10.0.1.9" x_forwarded_proto:"http" vcap_request_id:"9b302968-b59d-41d3-56a8-3c189d00e53a" response_time:20.628416 gorouter_time:20.628416 app_id:"ff858e3a-87ab-4566-8abb-534d8ac06e9e" app_index:"0" instance_id:"5e2a742c-6fe0-4ca5-6c7d-0492" x_cf_routererror:"-" x_b3_traceid:"3e86f4ce18bf70a02ac3d78b87092fbd" x_b3_spanid:"2ac3d78b87092fbd" x_b3_parentspanid:"-" b3:"3e86f4ce18bf70a02ac3d78b87092fbd-2ac3d78b87092fbd"
```

2. ✅ I validated that your update fixed the issue
```
2023-10-04T21:41:23.83+0000 [RTR/0] OUT go-responder.band-160711.cf-app.com - [2023-10-04T21:41:10.591502359Z] "GET /ws HTTP/1.1" 101 0 0 "-" "curl/7.81.0" "10.0.1.9:41256" "10.0.1.14:61002" x_forwarded_for:"10.0.1.9" x_forwarded_proto:"http" vcap_request_id:"15000039-c2fc-41c8-40cb-ea974f6dfc83" response_time:13.243427 gorouter_time:0.000428 app_id:"ff858e3a-87ab-4566-8abb-534d8ac06e9e" app_index:"0" instance_id:"5e2a742c-6fe0-4ca5-6c7d-0492" x_cf_routererror:"-" x_b3_traceid:"15000039c2fc41c840cbea974f6dfc83" x_b3_spanid:"40cbea974f6dfc83" x_b3_parentspanid:"-" b3:"15000039c2fc41c840cbea974f6dfc83-40cbea974f6dfc83"
```

3.  ✅ I validated that your update worked when `enable_log_attempts_details` is enabled.
```
2023-10-04T21:45:08.56+0000 [RTR/0] OUT go-responder.band-160711.cf-app.com - [2023-10-04T21:44:58.730948099Z] "GET /ws HTTP/1.1" 101 0 0 "-" "curl/7.81.0" "10.0.1.9:58760" "10.0.1.14:61002" x_forwarded_for:"10.0.1.9" x_forwarded_proto:"http" vcap_request_id:"f611f8ac-0d2a-4318-759d-d194c46b6170" response_time:9.837065 gorouter_time:0.000473 app_id:"ff858e3a-87ab-4566-8abb-534d8ac06e9e" app_index:"0" instance_id:"5e2a742c-6fe0-4ca5-6c7d-0492" failed_attempts:0 failed_attempts_time:"-" dns_time:"-" dial_time:0.001511 tls_time:0.005989 backend_time:9.836591 x_cf_routererror:"-" x_b3_traceid:"f611f8ac0d2a4318759dd194c46b6170" x_b3_spanid:"759dd194c46b6170" x_b3_parentspanid:"-" b3:"f611f8ac0d2a4318759dd194c46b6170-759dd194c46b6170"
```

Overall it would be great to remove as many sleeps as we can from the tests. But other than that, it looks good. Thank you for doing this important work!

domdom82:
hi @ameowlia thanks for the review! I've made all second sleeps into millisecond sleeps, saving about 5s of runtime for the proxy suite. I've also fixed some of the regular http access log tests and reduced their sleep times to 100ms each.

ameowlia (approved):

PR #367, opened 2023-09-25
Optional mTLS client certificate metadata verification

peanball:

We want to add optional mTLS client certificate metadata verification in addition to the issuer CA validity checks.

Given a particular signing CA (identified by its subject) we want to limit the allowed client certificate subjects that are allowed to send requests.

While the underlying issue is better solved with changes to the PKI and trust relationship between CAs, the PKI or its use by other sub-entities is not always under the control of the operator.

The use case is an upstream proxy managed by a different entity that uses a widely spread private CA for issuing its certificates.

We want to avoid that every user of that CA is able to issue requests to Gorouter, while still explicitly allowing the upstream proxy and its CA certificates, which are short-lived and need rotation.

In the specific use case, we have a CA that controls, which certificates are issued for specific sub-entities, which are clearly identified in the certificate's subject DN. 

e.g. configuration in routing-release:
```yaml
router:
  verify_client_certificate_metadata:
    - issuer_in_chain:
        common_name: Some Root CA
        country:
          - DE
      valid_cert_subjects:
        - common_name: Awesome Service 1
          organizational_unit:
            - sub-entity4
          organization:
            - Acme, Inc.
          country:
            - US
        - common_name: Awesome Service 2
          organizational_unit:
            - sub-entity7
          organization:
            - Acme, Inc.
          country:
            - US
```

1. Send a request using a client certificate that is signed by a CA that is configured in `client_ca_certs:` 
   but not in the explicit allowed subjects, configured in `verify_client_certificate_metadata:`:
   ```shell
   curl --cacert other-department-5.pem https://app.cf.example.com
   ```
2. Send a request using a client certificate that is signed by a CA that is configured in `client_ca_certs:`
   and in the allowed subjects, configured in `verify_client_certificate_metadata:`:
   ```shell
   curl --cacert awesome-service-2.pem https://app.cf.example.com
   ```

1. The request fails with a TLS handshake error. Gorouter logs the certificate subject and the fact that this certificate was not allowed.
2. succeeds, as desired.

1. The request passes through without TLS handshake errors, allows access to a different party that should not have access.
2. succeeds, as desired.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -213,9 +213,9 @@ type HTTPRewriteResponses struct {
 // ValidSubjects.
 type VerifyClientCertificateMetadataRule struct {
 	// The issuer DN , for which the subject validation should apply
-	CASubject CertSubject `yaml:"ca_subject"`
+	CASubject CertSubject `yaml:"issuer_in_chain"`
 	// The subject DNs	 that are allowed to be used for mTLS connections to Gorouter
-	ValidSubjects []CertSubject `yaml:"valid_subjects"`
+	ValidSubjects []CertSubject `yaml:"valid_cert_subjects"`
 }
 
 // CertSubject defines the same fields as pkix.Name and allows YAML declaration of said fields. This is used to
--- a/config/config.go
+++ b/config/config.go
@@ -222,8 +222,8 @@ type VerifyClientCertificateMetadataRule struct {
 // express distinguished names for certificate subjects in a comparable manner.
 type CertSubject struct {
 	Country            []string `yaml:"country"`
-	Organization       []string `yaml:"organisations"`
-	OrganizationalUnit []string `yaml:"organisation_units"`
+	Organization       []string `yaml:"organisation"`
+	OrganizationalUnit []string `yaml:"organisation_unit"`
 	CommonName         string   `yaml:"common_name"`
 	SerialNumber       string   `yaml:"serial_number"`
 	Locality           []string `yaml:"locality"`
@@ -291,20 +291,23 @@ func checkIfRuleAppliesToChain(chain []*x509.Certificate, logger logger.Logger, 
 // * the chain does not contain any client certificates (i.e. IsCA == false).
 func checkClientCertificateMetadataRule(chain []*x509.Certificate, logger logger.Logger, rule VerifyClientCertificateMetadataRule) error {
 	for _, cert := range chain {
-		if !cert.IsCA {
-			subject := cert.Subject
-			for _, validSubject := range rule.ValidSubjects {
-				vaildCertSubject := validSubject.ToName()
-				if vaildCertSubject.ToRDNSequence().String() == subject.ToRDNSequence().String() {
-					logger.Debug("chain", zap.String("issuer", cert.Issuer.String()), zap.Bool("CA", cert.IsCA), zap.String("subject", cert.Subject.String()))
-					return nil
-				}
-			}
-			logger.Warn("invalid-subject", zap.String("issuer", cert.Issuer.String()), zap.String("subject", cert.Subject.String()), zap.Object("allowed", rule.ValidSubjects))
-			return fmt.Errorf("subject not in the list of allowed subjects for CA Subject %q: %q", rule.CASubject, subject)
+		if cert.IsCA {
+			continue
 		}
+		subject := cert.Subject
+		for _, validSubject := range rule.ValidSubjects {
+			vaildCertSubject := validSubject.ToName()
+			if vaildCertSubject.ToRDNSequence().String() == subject.ToRDNSequence().String() {
+				logger.Debug("chain", zap.String("issuer", cert.Issuer.String()), zap.Bool("CA", cert.IsCA), zap.String("subject", cert.Subject.String()))
+				return nil
+			}
+		}
+		logger.Warn("invalid-subject", zap.String("issuer", cert.Issuer.String()), zap.String("subject", cert.Subject.String()), zap.Object("allowed", rule.ValidSubjects))
+		return fmt.Errorf("subject not in the list of allowed subjects for CA Subject %q: %q", rule.CASubject, subject)
 	}
-	return fmt.Errorf("incomplete chain")
+	// this should never happen as the function is only called on successful client certificate verification as callback
+	// to tls.Config.VerifyPeerCertificate.
+	return fmt.Errorf("cert chain provided to client certificate metadata verification did not contain a leaf certificate; this should never happen")
 }
 
 // InitClientCertMetadataRules compares the defined rules against client CAs set in `client_ca_certs`. When a rule
@@ -370,7 +373,8 @@ type Config struct {
 	ClientCertificateValidation                     tls.ClientAuthType                    `yaml:"-"`
 	OnlyTrustClientCACerts                          bool                                  `yaml:"only_trust_client_ca_certs"`
 	TLSHandshakeTimeout                             time.Duration                         `yaml:"tls_handshake_timeout"`
-	VerifyClientCertificateMetadata   []VerifyClientCertificateMetadataRule `yaml:"verify_client_certificate_metadata,omitempty"`
+	VerifyClientCertificatesBasedOnProvidedMetadata bool                                  `yaml:"enable_verify_client_certificate_metadata,omitempty"`
+	VerifyClientCertificateMetadataRules            []VerifyClientCertificateMetadataRule `yaml:"verify_client_certificate_metadata,omitempty"`
 
 	LoadBalancerHealthyThreshold    time.Duration `yaml:"load_balancer_healthy_threshold,omitempty"`
 	PublishStartMessageInterval     time.Duration `yaml:"publish_start_message_interval,omitempty"`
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -198,6 +203,130 @@ type HTTPRewriteResponses struct {
...
+			return fmt.Errorf("subject not in the list of allowed subjects for CA Subject %q: %q", rule.CASubject, subject)
+		}
+	}
+	return fmt.Errorf("incomplete chain")
+}
+
+// InitClientCertMetadataRules compares the defined rules against client CAs set in `client_ca_certs`. When a rule
+// is found that does not have a corresponding client CA (based on the CA's subject) that matches the rule, startup will fail.
+//
+// This is to avoid defining a rule with a minor typo that would then not apply at all and would make the whole
```
ameowlia, line 313: This makes sense that we would want to check for typos, but I worry that failing at runtime is inconvenient and late in the process. Is there a way to fail at templating time instead (or in addition)?

config/config.go
```diff
@@ -198,6 +203,130 @@ type HTTPRewriteResponses struct {
 	RemoveHeaders          []HeaderNameValue `yaml:"remove_headers,omitempty"`
 }
 
+// VerifyClientCertificateMetadata defines verification rules for client certificates, which allow additional checks
```
ameowlia, line 206: We don't typically have such verbose comments in runtime code, but given the development on this codebase is slowing down and is relatively stable, I think it is helpful and I like it.

config/config.go
```diff
@@ -688,6 +819,18 @@ func (c *Config) buildClientCertPool() error {
 		}
 	}
 	c.ClientCAPool = certPool
+
+	if c.VerifyClientCertificateMetadata != nil {
```
ameowlia, line 823: Instead of seeing if this property is nil, can you add a separate boolean property `VerifyClientCertificatesBasedOnProvidedMetadata` or something? I have run into issues in the past what I try to turn an obj for a feature into a boolean identifier, because then invariably, we add another property for that feature, and then things become a muddled mess.

config/config_test.go
```diff
@@ -1408,7 +1408,25 @@ route_services_secret_decrypt_only: 1PfbARmvIn6cgyKorA1rqR2d34rBOo+z3qJGz17pi8Y=
...
+					BeforeEach(func() {
+						configSnippet.ClientCACerts = string(clientRSAPEM)
+						configSnippet.VerifyClientCertificateMetadata = []VerifyClientCertificateMetadataRule{
+							{
+								CASubject:     CertSubject{CommonName: "abc.com"},
+								ValidSubjects: []CertSubject{CertSubject{Organization: []string{"abc, Inc."}}},
+							},
+						}
+					})
+					It("fails to validate", func() {
```
ameowlia, line 1421:
```suggestion
					It("returns a meaningful error", func() {
```

router/router.go
```diff
@@ -274,6 +279,17 @@ func (r *Router) serveHTTPS(server *http.Server, errChan chan error) error {
...
 }
 
+// verifyMtlsMetadata checks the Config.VerifyClientCertificateMetadata rules, if any are defined.
+//
+// Returns an error if one of the applicable verification rules fails.
+func (r *Router) verifyMtlsMetadata(_ [][]byte, chains [][]*x509.Certificate) error {
+	if chains != nil {
+		return config.VerifyClientCertMetadata(r.config.VerifyClientCertificateMetadata, chains, r.logger)
+	}
+	fmt.Printf("Chains length %d", len(chains))
```
ameowlia, line 289:
```suggestion
```

👻 You probably didn't mean to leave this in :)

router/router_test.go
```diff
@@ -1862,6 +1862,64 @@ var _ = Describe("Router", func() {
...
+								BeforeEach(func() {
+									config.VerifyClientCertificateMetadata = []cfg.VerifyClientCertificateMetadataRule{
+										{
+											CASubject:     cfg.CertSubject{CommonName: "abc.com"},
+											ValidSubjects: []cfg.CertSubject{cfg.CertSubject{Organization: []string{"abc, Inc."}}},
+										},
+									}
+								})
+								It("successfully serves SSL traffic", func() {
+
```
ameowlia, line 1918:
```suggestion
```

config/config.go
```diff
@@ -198,6 +203,130 @@ type HTTPRewriteResponses struct {
...
+				if vaildCertSubject.ToRDNSequence().String() == subject.ToRDNSequence().String() {
+					logger.Debug("chain", zap.String("issuer", cert.Issuer.String()), zap.Bool("CA", cert.IsCA), zap.String("subject", cert.Subject.String()))
+					return nil
+				}
+			}
+			logger.Warn("invalid-subject", zap.String("issuer", cert.Issuer.String()), zap.String("subject", cert.Subject.String()), zap.Object("allowed", rule.ValidSubjects))
+			return fmt.Errorf("subject not in the list of allowed subjects for CA Subject %q: %q", rule.CASubject, subject)
+		}
+	}
+	return fmt.Errorf("incomplete chain")
```
ameowlia, line 307: 1. Is there a test for this case?
2. Can you provide a more verbose error message? I don't think I would understand this message without some code digging to find your helpful comments.

router/router_test.go
```diff
@@ -1862,6 +1862,64 @@ var _ = Describe("Router", func() {
 						It("unsuccessfully serves SSL traffic", func() {
 							expectSuccessfulConnection()
 						})
+
+						Context("when verify_client_certificate_metadata is provided ", func() {
+							Context("when the client cert subject is one of the valid subjects for the specified CA subject", func() {
```
ameowlia, line 1867: What about a case where the cert has more info than is specified in the metadata?

For example...
1. The metadata says the country must be USA. 
1. But the cert includes country is USA, and company is Neopets. 

Hopefully it will succeed because the cert matches all of the restrictions explicitly specified in the metadata?

Can we test a case like this?

ameowlia (changes requested):

peanball:
@ameowlia thanks so much for this thoughtful and thorough review. We'll address the points. Fully agree on having a check in the spec ERB template / at template generation time.

router/router_test.go
peanball: That would actually not match. Internally we convert all of this into a [`pkix.Name`](https://pkg.go.dev/crypto/x509/pkix#Name) as you've likely seen.

The design is for an exact match at the moment, i.e. your configuration needs to be exhaustive.

There is one caveat regarding the matching of `pkix.Name`: We only use the base properties, not `Names` or `ExtraNames` at the moment.

But we can have a test nonetheless that would show exactly that, i.e. that it would not match. Any field that is not defined _must_ be empty right now.

config/config.go
peanball: No there is no test yet.

Yes we will add a test and a more verbose error message. It is possible/likely that this will not get through the regular client certificate check already. In that case I would remove it.

config/config.go
peanball: Yes, at least in addition. While it's unlikely that e.g. Gorouter would be used outside the routing-release, I would prefer to have a safeguard in the Gorouter code anyway.

ameowlia: in addition is great!

config/config.go
peanball: I have provided a more verbose error message and a comment.

I couldn't find a way to get to that point of having an incomplete chain processed by the callback for verifying the peer cert.

peanball:
@ameowlia Could you have another look?

I've updated this PR addressing all comments and provided the companion PR for routing release: https://github.com/cloudfoundry/routing-release/pull/355

config/config.go
```diff
@@ -198,6 +203,132 @@ type HTTPRewriteResponses struct {
...
+				vaildCertSubject := validSubject.ToName()
+				if vaildCertSubject.ToRDNSequence().String() == subject.ToRDNSequence().String() {
+					logger.Debug("chain", zap.String("issuer", cert.Issuer.String()), zap.Bool("CA", cert.IsCA), zap.String("subject", cert.Subject.String()))
+					return nil
+				}
+			}
+			logger.Warn("invalid-subject", zap.String("issuer", cert.Issuer.String()), zap.String("subject", cert.Subject.String()), zap.Object("allowed", rule.ValidSubjects))
+			return fmt.Errorf("subject not in the list of allowed subjects for CA Subject %q: %q", rule.CASubject, subject)
+		}
+	}
```
b1tamara, line 306:
```suggestion
	for _, cert := range chain {
		if cert.IsCA {
                     continue
		}
		subject := cert.Subject
		for _, validSubject := range rule.ValidSubjects {
			vaildCertSubject := validSubject.ToName()
			if vaildCertSubject.ToRDNSequence().String() == subject.ToRDNSequence().String() {
				logger.Debug("chain", zap.String("issuer", cert.Issuer.String()), zap.Bool("CA", cert.IsCA), zap.String("subject", cert.Subject.String()))
				return nil
			}
		}
		logger.Warn("invalid-subject", zap.String("issuer", cert.Issuer.String()), zap.String("subject", cert.Subject.String()), zap.Object("allowed", rule.ValidSubjects))
		return fmt.Errorf("subject not in the list of allowed subjects for CA Subject %q: %q", rule.CASubject, subject)	
	}
```

b1tamara: @peanball a small suggestion to keep the code simple

peanball: integrated. thank you!

ameowlia (approved):

PR #372, opened 2023-11-13
Fix nil panic

maxmoehl:
* A short explanation of the proposed change: Only call close on the original requests body if it is non-nil, otherwise this will cause a panic.

* An explanation of the use cases your change solves: Prevent panics.

* Expected result after the change: No panic if we have a client request without a body and the client disconnects early.

* Current result before the change: gorouter panics if we have a client request without a body and the client disconnects early.

* Links to any other associated PRs: #333

* [X] I have [run all the unit tests](https://github.com/cloudfoundry/routing-release#running-unit-and-integration-tests).


maxmoehl:
I'll try to come up with a regression test for this.

domdom82 (approved):

PR #375, opened 2023-12-01
Improve pool.Endpoint.Equals() peformance

peanball:

The change improves the `Endpoint.Equal()` function to use less time and no allocations during comparison. In profiler runs, @domdom82 identified that a lot of time is spent on the two `fmt.Sprintf` calls for comparison of the `tags` maps.

Route registration messages that happen on route refresh are compared with the existing endpoints in the route registry for every received `router.register` message. This happens for every route by default every 20 seconds. In our scenarios we have up to 70k routes (on each gorouter) and such comparisons add up.

Functionality is not changed. A benchmark has been added to show the changes.

Route registration (and re-registration) should be a little faster. This will have measurable effect for large numbers of routes and route registration messages only.

Route registration takes about 8% of the processing time in Gorouter in our profiling examples. See https://github.com/cloudfoundry/routing-release/issues/366.

### DIFF
--- a/registry/registry_benchmark_test.go
+++ b/registry/registry_benchmark_test.go
@@ -69,6 +69,7 @@ func BenchmarkRegisterWith100KRoutes(b *testing.B) {
 	for i := 0; i < b.N; i++ {
 		r.Register("foo50000.example.com", fooEndpoint)
 	}
+	b.ReportAllocs()
 }
 
 func BenchmarkRegisterWithOneRoute(b *testing.B) {
@@ -81,4 +82,5 @@ func BenchmarkRegisterWithOneRoute(b *testing.B) {
 	for i := 0; i < b.N; i++ {
 		r.Register("foo.example.com", fooEndpoint)
 	}
+	b.ReportAllocs()
 }
--- a/route/pool.go
+++ b/route/pool.go
@@ -3,6 +3,7 @@ package route
 import (
 	"encoding/json"
 	"fmt"
+	"maps"
 	"math/rand"
 	"net/http"
 	"sync"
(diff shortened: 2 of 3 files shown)


peanball:
Some benchmark results:

There is a new benchmark, `pool_bench_test.go`:
```
BenchmarkEndpointEquals-16           	  695605	      1597 ns/op	     480 B/op	      16 allocs/op <- fmt.Sprintf
BenchmarkEndpointEquals-16           	13326138	        83.46 ns/op	       0 B/op	       0 allocs/op <- maps.Equal
```

While this is a microbenchmark and the difference between those two ops looks very stark, the pool endpoint insertion is a small part of the overall processing. Eliminating allocations should be a significant benefit however for this often-called function in terms of cache coherency, reducing memory access, etc.

Finally, the results of `registry_benchmark_test.go`:
```
old version
BenchmarkRegisterWith100KRoutes-16    	  939470	      1712 ns/op	    1546 B/op	      11 allocs/op
BenchmarkRegisterWithOneRoute-16      	  977769	      1208 ns/op	    1546 B/op	      11 allocs/op

new version
BenchmarkRegisterWith100KRoutes-16    	 1000000	      1312 ns/op	    1440 B/op	       7 allocs/op
BenchmarkRegisterWithOneRoute-16      	 1281951	       839.3 ns/op	    1440 B/op	       7 allocs/op
```

There are fewer allocations and generally less time per iteration with the new comparator.

The semantics of `Endpoint.Equals()` remain identical.

maxmoehl:
When merging this into routing-release, please remember to bump [the `go` directive](https://github.com/cloudfoundry/routing-release/blob/69dba350cbc5d2a92ae3d5b15309d3b6899c0814/src/code.cloudfoundry.org/go.mod#L3) to at least `1.21.0`.

ameowlia (approved):
Nice! Great find!

ameowlia:
https://github.com/cloudfoundry/routing-release/commit/da3bb317a517178a86575bc8ff0e6b8601504532 - merged into routing release

ameowlia:
Oh no! This is failing go vet...

```
go vet ./...
# code.cloudfoundry.org/gorouter/route_test
gorouter/route/pool_bench_test.go:35:14: variable declaration copies lock value to endpoint3: code.cloudfoundry.org/gorouter/route.Endpoint contains sync.RWMutex
```

domdom82:
Neat!

PR #378, opened 2023-12-10
Optimistic route registry locks

peanball:

This PR reduces the time spent in exclusive write locks for the RouteRegistry.

A lot of the work that is currently done with an exclusive write lock is to look up and determine, _whether_ a change is necessary at all. This PR carries out such lookup tasks with a read lock and acquires the write lock only when necessary.

The main issue with write locking is that this impacts parallel execution, e.g. of Gorouter requests. This is particularly prevalent when a large number of routes and their corresponding route.register messages are arriving at each Gorouter.

Having the lookup and decision part, whether a change is necessary under a read lock, allows other routines (e.g. serving requests) to be executed in parallel.

As Gorouter is ultimately an "eventually consistent" system, there is no hard requirement for sequential processing. While it is desirable that a route register or deregister message is processed as quickly as possible, there is no hard requirement for "correct" sequence of execution. Using a message bus and multiple participants in registering routes, the order, arrival or delay/latency of information propagation are not guaranteed.

By relaxing and decoupling the "do I need to change anything" from regular work and look-ups improves the 95% case of Gorouter's work, while not impacting route registry functionality.

This is a "drop in" performance improvement. Gorouter is to be operated as before, but should be able to do more work concurrently, especially with large numbers of routes in the route registry update cycle.

A benchmark was added in addition to the change. This benchmark measures the throughput of `Register` in the route registry, while concurrently there are 10 goroutines that do `Lookup` as quickly as possible in parallel.

This simulates the regular Gorouter case of having to look up endpoints to proxy requests to a target, while route register messages are coming in, in parallel.

The go benchmarking tool anneals to an optimal benchmarking iteration size. This can be seen by the outputs of different log messages of the lookup goroutines running in parallel. Only the last output of those is relevant for the displayed timing.

Please also note that the time and memory metrics of the benchmark include the concurrently running goroutines. As a lookup is (currently) requiring an allocation, the more lookups are done, the more allocations will still show up in the benchmark result. These allocations are _not_ part of the `Register` workload that is measured.

Benchmark result:
```
goos: linux
goarch: amd64
pkg: code.cloudfoundry.org/gorouter/registry
cpu: Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz
BenchmarkRegisterWith100KRoutes-4                       	  545870	      2217 ns/op	    1440 B/op	       7 allocs/op
BenchmarkRegisterWithOneRoute-4                         	  915207	      1550 ns/op	    1440 B/op	       7 allocs/op
BenchmarkRegisterWithConcurrentLookupWith100kRoutes-4   	  628071	     10911 ns/op	    2525 B/op	      40 allocs/op
--- BENCH: BenchmarkRegisterWithConcurrentLookupWith100kRoutes-4
    registry_benchmark_test.go:139: Looked up 0 routes concurrently, registered 1
    registry_benchmark_test.go:139: Looked up 1011 routes concurrently, registered 100
    registry_benchmark_test.go:139: Looked up 70499 routes concurrently, registered 10000
    registry_benchmark_test.go:139: Looked up 21307628 routes concurrently, registered 628071
PASS
ok  	code.cloudfoundry.org/gorouter/registry	34.591s
```

When applying only the commit with the Benchmark and the existing implementation:
```
goos: linux
goarch: amd64
pkg: code.cloudfoundry.org/gorouter/registry
cpu: Intel(R) Xeon(R) Platinum 8259CL CPU @ 2.50GHz
BenchmarkRegisterWith100KRoutes-4                       	  623066	      2097 ns/op	    1440 B/op	       7 allocs/op
BenchmarkRegisterWithOneRoute-4                         	  858085	      1388 ns/op	    1440 B/op	       7 allocs/op
BenchmarkRegisterWithConcurrentLookupWith100kRoutes-4   	   94538	     18422 ns/op	    1904 B/op	      21 allocs/op
--- BENCH: BenchmarkRegisterWithConcurrentLookupWith100kRoutes-4
    registry_benchmark_test.go:139: Looked up 0 routes concurrently, registered 1
    registry_benchmark_test.go:139: Looked up 1993 routes concurrently, registered 100
    registry_benchmark_test.go:139: Looked up 131009 routes concurrently, registered 10000
    registry_benchmark_test.go:139: Looked up 1373364 routes concurrently, registered 94538
PASS
ok  	code.cloudfoundry.org/gorouter/registry	29.733s
```

As stated above, only the last output line is the "maximum" achieved. The other lines are the warm-up and calibration done by the Go benchmarking framework.

Comparison:
| Metric |`Register` | `Lookup`* |
| -------|---------:|--------:|
| Existing  | `94538` |  `1373364` |
| This PR | `628071` | `21307628 |

\* (sum of 10 concurrent goroutines)

**Please note**, this is a micro-benchmark, so the values may fluctuate between runs depending on other load, and depending on the type of machine used. Route registration is only part of Gorouter's work. The improvement in reduced concurrent lock times and resulting throughput is clearly visible in either case, however.

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -49,6 +49,7 @@ type RouteRegistry struct {
 
 	ticker           *time.Ticker
 	timeOfLastUpdate time.Time
+	updateTimeLock   sync.RWMutex
 
 	routingTableShardingMode string
 	isolationSegments        []string
@@ -103,13 +104,37 @@ func (r *RouteRegistry) Register(uri route.Uri, endpoint *route.Endpoint) {
 }
 
 func (r *RouteRegistry) register(uri route.Uri, endpoint *route.Endpoint) route.PoolPutResult {
+	r.RLock()
+	defer r.RUnlock()
+
+	t := time.Now()
+	routekey := uri.RouteKey()
+	pool := r.byURI.Find(routekey)
+
+	if pool == nil {
+		// release read lock, insertRouteKey() will acquire a write lock.
+		r.RUnlock()
+		pool = r.insertRouteKey(routekey, uri)
+		r.RLock()
+	}
+
+	if endpoint.StaleThreshold > r.dropletStaleThreshold || endpoint.StaleThreshold == 0 {
+		endpoint.StaleThreshold = r.dropletStaleThreshold
+	}
+
+	endpointAdded := pool.Put(endpoint)
+
+	r.SetTimeOfLastUpdate(t)
+
+	return endpointAdded
+}
+
+// insertRouteKey acquires a write lock, inserts the route key into the registry and releases the write lock.
+func (r *RouteRegistry) insertRouteKey(routekey route.Uri, uri route.Uri) *route.EndpointPool {
 	r.Lock()
 	defer r.Unlock()
 
-	t := time.Now()
-
-	routekey := uri.RouteKey()
-
+	// double check that the route key is still not found, now with the write lock.
 	pool := r.byURI.Find(routekey)
 	if pool == nil {
 		host, contextPath := splitHostAndContextPath(uri)
@@ -125,16 +150,7 @@ func (r *RouteRegistry) register(uri route.Uri, endpoint *route.Endpoint) route.
 		// for backward compatibility:
 		r.logger.Debug("uri-added", zap.Stringer("uri", routekey))
 	}
-
-	if endpoint.StaleThreshold > r.dropletStaleThreshold || endpoint.StaleThreshold == 0 {
-		endpoint.StaleThreshold = r.dropletStaleThreshold
-	}
-
-	endpointAdded := pool.Put(endpoint)
-
-	r.timeOfLastUpdate = t
-
-	return endpointAdded
+	return pool
 }
 
 func (r *RouteRegistry) Unregister(uri route.Uri, endpoint *route.Endpoint) {
@@ -281,6 +297,8 @@ func (registry *RouteRegistry) NumUris() int {
 }
 
 func (r *RouteRegistry) MSSinceLastUpdate() int64 {
+	r.RLock()
+	defer r.RUnlock()
 	timeOfLastUpdate := r.TimeOfLastUpdate()
 	if (timeOfLastUpdate == time.Time{}) {
 		return -1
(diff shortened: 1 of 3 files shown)


registry/registry.go
```diff
@@ -103,13 +104,37 @@ func (r *RouteRegistry) Register(uri route.Uri, endpoint *route.Endpoint) {
...
+
+	return endpointAdded
+}
+
+// insertRouteKey acquires a write lock, inserts the route key into the registry and releases the write lock.
+func (r *RouteRegistry) insertRouteKey(routekey route.Uri, uri route.Uri) *route.EndpointPool {
+	r.Lock()
+	defer r.Unlock()
 
+	// double check that the route key is still not found, now with the write lock.
```
domdom82, line 137: the function call to `r.byURI.Find(routekey)` is pretty expensive as it traverses through the Trie structure sequentially. If possible I'd like to avoid doing that twice. In what scenario could there have been a change to the Trie between here and L112? While you're doing the first lookup you already have a read-lock obtained, so no one else could have acquired a write-lock and have inserted an element at the `routekey` location that would conflict? so why do the lookup twice?

domdom82 (changes requested):

registry/registry.go
peanball: It's an instance of [Double Checked Locking](https://en.wikipedia.org/wiki/Double-checked_locking).

A process or thread can be interrupted/preempted at pretty much any time, so there is a slim chance that we look up reading, release the lock, and before _we_ acquire the write lock some other thread may have acquired it and changed the value. And _then_ we get the write lock and work on outdated assumptions.

Without this double check, there is a low-probability race condition that this double check avoids entirely.

The lookup of `r.byURI.Find()` is reasonably trivial. The route URI is split by `/` and those segments are iterated over.
This is actually also improved further by #379, by eliminating the alloc caused by the `parts()` function.

domdom82: > eliminating the alloc caused by the parts() function

I was worried about that exact part 😁
Thanks for pointing to the other fix in that area.

domdom82 (approved):

peanball:
@domdom82, it already runs with the race detector and had found some races in earlier development that are now fixed.

I was very happy to see the race detector in action :)

PR #379, opened 2023-12-10
perf(route-registry): avoid heap alloc on uri split

peanball:

`string.SplitN()` will allocate the result array for the results. When splitting into two parts, this can be done without allocating a slice of results dynamically, as it will be up to 2 results.

The `strings` package with the function `Cut` that does exactly that.

Adapted code in:
* The URI mapping in the Trie used for the Route Registry
* The route service FQDN stripper function
* Route registry host and context path splitter
* W3C Tracestate parser (splits `key=value` into `key`, `value`)

Fewer allocations, especially on lookups, help with performance. Heap allocations are expensive.

No change for operators.

Fewer allocs / operation, e.g. when using the route registry trie.

```
goos: darwin
goarch: amd64
pkg: code.cloudfoundry.org/gorouter/registry
cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
BenchmarkRegisterWith100KRoutes
BenchmarkRegisterWith100KRoutes-16    	  950095	      2599 ns/op	    1408 B/op	       6 allocs/op
BenchmarkRegisterWithOneRoute
BenchmarkRegisterWithOneRoute-16      	 1179136	       936.1 ns/op	    1408 B/op	       6 allocs/op
PASS
```

Please ignore the op times. The results were acquired under different loads (test suite running in the background in Docker). Only the 6 allocs/op (vs 7 below) is relevant to highlight.

as this is an improvement, the current state is the baseline:

```
goos: darwin
goarch: amd64
pkg: code.cloudfoundry.org/gorouter/registry
cpu: Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
BenchmarkRegisterWith100KRoutes
BenchmarkRegisterWith100KRoutes-16    	  725313	      1684 ns/op	    1440 B/op	       7 allocs/op
BenchmarkRegisterWithOneRoute
BenchmarkRegisterWithOneRoute-16      	  700059	      1766 ns/op	    1440 B/op	       7 allocs/op
```


registry/container/trie.go
```diff
@@ -51,8 +51,8 @@ func (r *Trie) MatchUri(uri route.Uri) *route.EndpointPool {
 	var lastPool *route.EndpointPool
 
 	for {
-		pathParts := parts(key)
-		SegmentValue := pathParts[0]
+		before, after, found := strings.Cut(key, "/")
```
maxmoehl, line 54: Do we need the new variables? `before` can simply become `segmentValue` (notice: lower-case :smile:). I'm not so sure for `key` but it looks like we could also assign that one directly (but please double-check that).

Bonus points for not allocating the remaining variables per-run but allocating them before and re-using them in each run.

This applies to all similar for-loops.

maxmoehl (commented):
Some minor optimisations :)

In general I'd like to see more descriptive names for the variables.

registry/container/trie.go
peanball: Fair point. Thanks. I will have another pass.

peanball: Done. Please have another look.

maxmoehl: That's the bonus point: allocating in the loop should be avoided entirely if possible.

maxmoehl: We agreed that this is beyond the scope of this PR.

maxmoehl (approved):

registry/container/trie.go
domdom82: maybe nits, but the `parts` func was probably introduced to avoid code duplication which is now reintroduced.
if you must remove it for performance reasons, I'd opt for making the "/" at least a const like `const pathSeparator = "/"` or something.

domdom82: Agreed to keep it the way it is and push the makeover to a later refactoring.

domdom82 (approved):

registry/container/trie.go
peanball: For completeness' sake a quick summary of what we discussed:
1. the `"/"` literal is strewn across the rest of the code. Pulling it out only in the `Trie` would not help significantly.
2. The `parts` function would then look as follows:
```go
func parts(key string) (before, after, found) {
    return strings.Cut(key, "/")
}
```
This does not bring much clarity to what is actually done. Because the whole loop and lookup logic is repeated 5 times, reducing repetition in this one spot will not help much, but adds yet another indirection for the reader of the code.
3. We agree that the `Trie` implementation is 70% repetition of almost the same code and should eventually be refactored. This PR does not aim to do that, just to remove the unnecessary allocation.

PR #380, opened 2023-12-13
Experiment: `http.ResponseController.EnableFullDuplex()`

peanball:
Experiments with `http.ResponseController.EnableFullDuplex()`.

### DIFF
--- a/integration/retry_test.go
+++ b/integration/retry_test.go
@@ -109,25 +109,16 @@ var _ = Describe("Retries", func() {
 			badApp.SetHandlers(handlers)
 			badApp.Listen()
 
-			successSeen := false
-			// we need to retry the entire http request many times to get a success until https://github.com/golang/go/issues/31259
-			// is resolved.
-			for i := 0; i < 100; i++ {
-				req := testState.newPostRequest(
-					fmt.Sprintf("http://%s", appURL),
-					bytes.NewReader([]byte(payload)),
-				)
-				resp, err := testState.client.Do(req)
-				Expect(err).NotTo(HaveOccurred())
-				if resp.StatusCode == http.StatusOK {
-					successSeen = true
-					break
-				}
-				if resp.Body != nil {
-					resp.Body.Close()
-				}
+			req := testState.newPostRequest(
+				fmt.Sprintf("http://%s", appURL),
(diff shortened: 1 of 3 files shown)


mariash:
EnableFullDuplex is currently failing on the ResponseController since ResponseWriter that is received in ServeHTTP is Negroni.ResponseWriter which does not provide EnableFullDuplex. There is an [issue](https://github.com/cloudfoundry/routing-release/issues/373) opened about Negroni being outdated.

peanball:
Yes, I've opened that issue after attempting with this branch. The PR was to share progress with @geofffranks and can be closed now. I will do that latest next week when I'm back at work.

peanball:
Closing this PR. It was an experiment with `EnableFullDuplex` that didn't work and resulted in https://github.com/cloudfoundry/routing-release/issues/373.

mariash:
@peanball I am looking into this right now. I think I would make PR to negroni to add this.

peanball:
Negroni looks like a dead project through.

Before making a PR there we might consider a locally patched version, try if it gives us the gains that @geofffranks was hoping for and only then consider resurrecting this project. 

That said, there might be other features we might like from newer Go versions now or later that would have to be added by us to Negroni instead of a framework's maintainers.

mariash:
@peanball so I looked into this more and negroni provides an `Unwrap` method to the ResponseWriter - https://github.com/urfave/negroni/commit/59d32aab419da1606c459c848c7586ef52b07a4b which will satisfy the latest ResponseController and eventually `EnableFullDuplex` will be called on the underlying http.ResponseWriter. 

Negroni actually has v2 and v3 versions, that are not visible on their releases page, but can be imported as `github.com/urfave/negroni/v3`. The problem is that `Unwrap` was not released on v3 yet. And looking at [Issue](https://github.com/urfave/negroni/issues/274) the last comment from December 2023 stated that author is temporarily unavailable. So I would give it couple more months to see if he will be back and cut a new release. For now I pinned the negroni to latest v3.

Here is the proposed change for the Gorouter: https://github.com/cloudfoundry/gorouter/pull/395
Related change in routing-release: https://github.com/cloudfoundry/routing-release/pull/389

PR #387, opened 2024-01-22
Fix ability to start gorouter with the default config

geofffranks:
- No longer require health listtener TLS cert configs when launching gorouter with a default config in test-mode
- Throw errors in config parsing + router logic if neither http nor https health listeners are enabled.
- Fix test names and bad URL mgmt for health listener tests.

routing-release requires the TLS certificate to be set still. cf-deployment generates this and passes it in. This change should only affect testing of gorouter.

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -144,7 +144,7 @@ func NewRouter(
 	}
 
 	if healthListener == nil && component == nil && healthTLSListener == nil {
-		return nil, fmt.Errorf("No TLS certificates provided and non-tls health listner disabled. No health listener can start. This is a bug in gorouter. This error should have been caught when parsing the config")
+		return nil, fmt.Errorf("No TLS certificates provided and non-tls health listener disabled. No health listener can start. This is a bug in gorouter. This error should have been caught when parsing the config")
 	}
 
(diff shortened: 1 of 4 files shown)


router/router.go
```diff
@@ -126,25 +126,27 @@ func NewRouter(
...
+			},
+			HealthCheck: healthCheck,
+		}
 		if err := healthTLSListener.ListenAndServe(); err != nil {
 			return nil, err
 		}
 	}
 
+	if healthListener == nil && component == nil && healthTLSListener == nil {
+		return nil, fmt.Errorf("No TLS certificates provided and non-tls health listner disabled. No health listener can start. This is a bug in gorouter. This error should have been caught when parsing the config")
```
domdom82, line 147:
```suggestion
		return nil, fmt.Errorf("No TLS certificates provided and non-tls health listener disabled. No health listener can start. This is a bug in gorouter. This error should have been caught when parsing the config")
```

domdom82 (commented):
found a typo. tested it locally and works fine. rest looks LGTM

domdom82 (approved):

PR #396, opened 2024-02-20
feat: hack: support the "Partitioned" flag on VCAP_ID cookies

peanball:

The `Partitioned` flag is used for cookies that are set on web sites embedded via iframes. The cookie is then available only in combination of the host site and the embedded site.

Golang's `http.Cookie` type does not yet support the `Partitioned` flag, but Google Chrome is already testing mandatory support and rejecting/omitting cookies without it for 1% of users via A/B roll-out.

The implementation wraps the `http.Cookie` and extends it with the `Partitioned` field. This field is then used to forward the raw cookie string when creating the derived VCAP_ID cookies for sticky sessions.

Once the Golang standard library supports the `Partitioned` flag, this wrapper can just be removed.

A test that checks the `Unparsed` section of the `http.Cookie` will ensure that the tests will fail once the `Partitioned` flag is supported by the Golang standard library.

Users of Safari and Google Chrome (currently 1% via A/B test, soon 100% of users) and websites that use iframes.

### DIFF
--- a/proxy/round_tripper/proxy_round_tripper_test.go
+++ b/proxy/round_tripper/proxy_round_tripper_test.go
@@ -993,7 +993,7 @@ var _ = Describe("ProxyRoundTripper", func() {
 
 			Context("when using sticky sessions", func() {
 				var (
-					sessionCookie *http.Cookie
+					sessionCookie *round_tripper.Cookie
 					endpoint1     *route.Endpoint
 					endpoint2     *route.Endpoint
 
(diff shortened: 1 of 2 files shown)


proxy/round_tripper/proxy_round_tripper_test.go
```diff
@@ -1365,10 +1370,14 @@ var _ = Describe("ProxyRoundTripper", func() {
 								newCookies := resp.Cookies()
 								Expect(newCookies).To(HaveLen(2))
 								Expect(newCookies[0].Raw).To(Equal(sessionCookie.String()))
+
+								// This should fail when Golang introduces parsing for the Partitioned flag on cookies.
```
maxmoehl, line 1374: Please also link https://github.com/golang/go/issues/62490 here.

maxmoehl (approved):
Change looks good! Just one minor comment.

proxy/round_tripper/proxy_round_tripper_test.go
peanball: Good idea, thanks!
Done. I've added the link to the implementation and the test.

PR #397, opened 2024-02-23
fix: Don't retry more often than endpoints available

domdom82:

Fixes [issue 385](https://github.com/cloudfoundry/routing-release/issues/385)

- Remove unlimited retries as it is dangerous. [companion PR for spec on routing-release](https://github.com/cloudfoundry/routing-release/pull/390)
- Minimum number of attempts is 1
- Do not attempt more often than endpoints available on the route
- Fixed test suite to actually use the number of backends as claimed in the tests (instead of trying 5 times on one endpoint, we will now actually try on 5 different endpoints)

With [a change](https://github.com/cloudfoundry/gorouter/commit/b820e884e3d11755e306ca2e14e83c207855091a) to error classifications, errors such as "IdempotentEOF" are no longer prunable. They only are retry-able now (which is correct, because we don't know for sure if the endpoint is unusable and must be pruned).

This has the side-effect that apps that are slow and buggy tend to exhaust `config.backends.maxAttempts`, which can be dangerous if `maxAttempts` is set to a high value or even unlimited, which could cause Gorouter to take extremely long to respond. This is because prior to this PR, Gorouter would have only stopped trying once `maxAttempts` had been reached, regardless whether or not all of the endpoints were non-working.

This PR makes it so that Gorouter will at most try the number of endpoints in the route before giving up. If there is only one endpoint and it's broken, there is no point in trying it 5 times in a row, it just slows everything down.

1. Deploy an app with 3 endpoints, all of them unresponsive
2. Set `config.backends.maxAttempts` to 5
3. Curl the app

Gorouter tries 3 times until all endpoints are exhausted

Gorouter tries 5 times, retrying endoints it already tried and failed

[routing-release PR](https://github.com/cloudfoundry/routing-release/pull/390)

### DIFF
--- a/proxy/round_tripper/proxy_round_tripper.go
+++ b/proxy/round_tripper/proxy_round_tripper.go
@@ -163,12 +163,12 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
 	var selectEndpointErr error
 	var maxAttempts int
 	if reqInfo.RouteServiceURL == nil {
-		maxAttempts = rt.config.Backends.MaxAttempts
+		maxAttempts = max(min(rt.config.Backends.MaxAttempts, reqInfo.RoutePool.NumEndpoints()), 1)
 	} else {
 		maxAttempts = rt.config.RouteServiceConfig.MaxAttempts
 	}
 
-	for attempt := 1; attempt <= maxAttempts || maxAttempts == 0; attempt++ {
+	for attempt := 1; attempt <= maxAttempts; attempt++ {
 		logger := rt.logger
 
 		// Reset the trace to prepare for new times and prevent old data from polluting our results.
--- a/proxy/round_tripper/proxy_round_tripper_test.go
+++ b/proxy/round_tripper/proxy_round_tripper_test.go
@@ -89,7 +89,8 @@ var _ = Describe("ProxyRoundTripper", func() {
 
 			reqInfo *handlers.RequestInfo
 
-			endpoint *route.Endpoint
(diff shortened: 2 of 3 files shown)


proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -163,12 +163,12 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
...
 	var maxAttempts int
 	if reqInfo.RouteServiceURL == nil {
-		maxAttempts = rt.config.Backends.MaxAttempts
+		maxAttempts = max(min(rt.config.Backends.MaxAttempts, reqInfo.RoutePool.NumEndpoints()), 1)
 	} else {
 		maxAttempts = rt.config.RouteServiceConfig.MaxAttempts
 	}
 
-	for attempt := 1; attempt <= maxAttempts || maxAttempts == 0; attempt++ {
+	for attempt := 1; attempt <= maxAttempts; attempt++ {
```
maxmoehl, line 171: This is a breaking change, do we know of anyone relying on this "unlimited" option?

domdom82: yes. see https://github.com/cloudfoundry/routing-release/pull/390

maxmoehl: Okay but the option was always a bit dangerous to begin with, given that there are retrieable errors which do not cause the endpoint to be pruned 😄

maxmoehl (approved):

PR #403, opened 2024-03-16
Remove obsolete request handler

domdom82:

OK, I know. It's quite a handful, this pull request. But hear me out:

While I was working on [the missing session affinity for web sockets](https://github.com/cloudfoundry/routing-release/issues/351) where I wanted to add a proper middleware for handling session cookies, I realized that this wouldn't work for web sockets. Why? The main reason is that the [request handler](https://github.com/cloudfoundry/gorouter/blob/main/proxy/proxy.go#L270) does not play by the rules of Negroni and does not call [the next function](https://github.com/cloudfoundry/gorouter/blob/main/proxy/proxy.go#L274), instead it just returns after handling the web socket.

This made it impossible to call any other middleware after request handler, which also meant that I couldn't use it to set session cookies.

While digging into the history of the request handler, I realized that Go itself already handles web sockets in [httputil.ReverseProxy](https://github.com/golang/go/commit/ee55f0856a3f1fed5d8c15af54c40e4799c2d32f) since Go 1.12. So there was no real reason for keeping the web socket handling in Gorouter.

There is a second issue with handling web sockets (or any protocol for that matter) outside the stdlib: RFC compliance.
While refactoring, I stumbled across a [test](https://github.com/cloudfoundry/gorouter/blob/main/integration/tls_to_backends_test.go#L84) that violates the web socket RFC: If the server does not respond with a 101 status on upgrade, the proxy is supposed to treat the response as regular HTTP, not as an error. However, request handler does and thus breaks the protocol compliance. It is one instance where implementing a standard protocol yourself can backfire.

Another weirdness that was refactored in this PR is the [Hop-By-Hop header filtering](https://github.com/cloudfoundry/gorouter/blob/main/proxy/handler/request_handler.go#L171) which is a generic HTTP feature that should belong in its own middleware, not a specific handler for web sockets. So I moved that piece into its proper handler, further improving the code structure.

The long-term benefit I am trying to achieve with this PR:
- Consolidation of the code base (remove large chunks of legacy code, including the entire `handler` package)
- Better standards compliance by relying on the Go stdlib wherever possible
- Streamline handling of requests (everything goes through ProxyRoundTripper now)
- Make it easier to add new functionality such as session handling via middleware

This PR lays the ground work for a follow-up PR that enables moving session affinity handling into a proper middleware, enabling it to work for both HTTP and web sockets requests, thus solving [the missing session affinity for web sockets](https://github.com/cloudfoundry/routing-release/issues/351)

This is a major refactoring effort. The behavior should be unchanged, except where RFC 6455 was violated by request handler before.

- All tests should pass

- All tests should pass


domdom82:
side note about routing-release:
- once this PR is merged, routing-release needs to remove [this line](https://github.com/cloudfoundry/routing-release/blob/develop/packages/gorouter/spec#L48) else the gosub spec tests will fail

handlers/hop_by_hop.go
maxmoehl, line 1: Just to be sure: this was completely (or at least mostly) copied from the request handler?

My only complaint is that the name doesn't feel quite right, it doesn't include the actual functionality of the handler which is filtering. Maybe `FilterConnectionHeader` is more appropriate?

proxy/proxy_test.go
```diff
@@ -2444,12 +2444,12 @@ var _ = Describe("Proxy", func() {
 					req, err := http.ReadRequest(conn.Reader)
 					Expect(err).NotTo(HaveOccurred())
 
-					done <- req.Header.Get("Upgrade") == "WebsockeT" &&
-						req.Header.Get("Connection") == "UpgradE"
+					done <- req.Header.Get("Upgrade") == "Websocket" &&
+						req.Header.Get("Connection") == "Upgrade"
```
maxmoehl, line 2448: Why did you adjust these tests?

handlers/hop_by_hop.go
domdom82: yes, this was moved from the RequestHandler where it didn't belong. I'm open for suggestions on the name. I started with hop-by-hop because it focuses on the `HopByHopHeadersToFilter` config property.

proxy/proxy_test.go
domdom82: The httputil.ReverseProxy takes care of sanitizing the hop-by-hop headers, too. After removing them according to [RFC 7230](https://github.com/golang/go/blob/master/src/net/http/httputil/reverseproxy.go#L294) it re-adds the headers necessary for the websocket update [here](https://github.com/golang/go/blob/master/src/net/http/httputil/reverseproxy.go#L413) but it adds them like a sane person would, so "UpdgraE" sent by your client becomes "Upgrade" at the other end.

The tests adjusted make the wrong assumptions that these headers are retained as-is and are not touched by the proxy - even though they are hop-by-hop headers that are supposed to be consumed by proxies.

handlers/hop_by_hop.go
maxmoehl: I didn't realise that the property is called like this. In that case it probably makes sense to stick with the name.

proxy/proxy.go
```diff
@@ -298,10 +260,16 @@ func (p *proxy) setupProxyRequest(target *http.Request) {
...
 	target.URL.RawQuery = ""
 
-	handler.SetRequestXRequestStart(target)
+	setRequestXRequestStart(target)
 	target.Header.Del(router_http.CfAppInstance)
 }
 
+func setRequestXRequestStart(request *http.Request) {
+	if _, ok := request.Header[http.CanonicalHeaderKey("X-Request-Start")]; !ok {
+		request.Header.Set("X-Request-Start", strconv.FormatInt(time.Now().UnixNano()/1e6, 10))
```
peanball, line 269: would [`time.Now().UnixMilli()`](https://cs.opensource.google/go/go/+/refs/tags/go1.22.1:src/time/time.go;l=1222) not be equivalent and more readable in terms of what this timestamp actually represents?

domdom82: probably. but that's outside the scope of this PR. I've just moved the code needed as-is from the request handler to proxy.

integration/gdpr_test.go
```diff
@@ -161,7 +167,9 @@ var _ = Describe("GDPR", func() {
 
 			x.Close()
 
-			Eventually(gbytes.BufferReader(testState.gorouterSession.Out)).Should(gbytes.Say(`"RemoteAddr":"-"`))
+			Eventually(func() ([]byte, error) {
+				return os.ReadFile(testState.AccessLogFilePath())
+			}).Should(ContainSubstring(`"foo-agent" "-"`))
```
geofffranks, line 172: Shouldn't this case be testing for `"RemoteAddr":"-"`?

integration/tls_to_backends_test.go
```diff
@@ -81,8 +81,13 @@ var _ = Describe("TLS to backends", func() {
...
 		})
 
-		It("closes connections with backends that respond with non 101-status code", func() {
-			wsApp := test.NewHangingWebSocketApp([]route.Uri{"ws-app." + test_util.LocalhostDNS}, testState.cfg.Port, testState.mbusClient, "")
+		// this test mandates RFC 6455 - https://datatracker.ietf.org/doc/html/rfc6455#section-4
+		// where it is stated that:
+		// "(...) If the status code received from the server is not 101, the
+		//       client handles the response per HTTP [RFC2616] procedures."
+		// Which means the proxy must treat non-101 responses as regular HTTP [ and not close the connection per se ]
+		It("does not close connections with backends that respond with non 101-status code", func() {
```
geofffranks, line 89: So, previously, gorouter behaved as such (RFC noncompliant):
- client requests a websocket upgrade
- server responds with a  status code indicating that no upgrade will happen
- gorouter closes the client connection immediately
- the client would need to create a new connection to make a subsequent request
- the client would not see any body data returned from gorouter (but headers would have been sent)

and will now behave like this (RFC compliant):
- client requests a websocket upgrade
- server responds with a  status code indicating that no upgrade will happen
- the client receives headers + body of the response
- the connection is kept-alive for re-use

Is that accurate?

handlers/hop_by_hop.go
```diff
@@ -0,0 +1,54 @@
...
+	connections := r.Header.Values("Connection")
+	for index, connection := range connections {
+		if connection != "" {
+			values := strings.Split(connection, ",")
+			connectionHeader := []string{}
+			for i := range values {
+				trimmedValue := strings.TrimSpace(values[i])
+				found := false
+				for _, item := range h.cfg.HopByHopHeadersToFilter {
+					if strings.ToLower(item) == strings.ToLower(trimmedValue) {
```
geofffranks, line 42: Staticcheck is complaining this, since `strings.EqualFold()` is a better case-insensitive string-equality checker for unicode text. Can you make that change?

integration/tls_to_backends_test.go
```diff
@@ -104,18 +109,12 @@ var _ = Describe("TLS to backends", func() {
...
 
-			// client-side conn should have been closed
-			// we verify this by trying to read from it, and checking that
-			//  - the read does not block
-			//  - the read returns no data
-			//  - the read returns an error EOF
-			n, err := conn.Read(make([]byte, 100))
-			Expect(n).To(Equal(0))
-			Expect(err).To(Equal(io.EOF))
+			data, err := io.ReadAll(resp.Body)
```
geofffranks, line 114: The value of `err` here is never checked. Can you either check it, or assign to `_`?

geofffranks (changes requested):
Added a couple questions + items to fix. 

Also it would probably be best to discuss the potential risks + benefits of this at the WG meeting in April, to see what others think. I'm in favor of making this change though.

integration/tls_to_backends_test.go
domdom82: Yes, that is correct. The main reason for this behavior is potential authentication requests by server indicated by a 401 and possible redirects via 302 / 301 to another endpoint that actually does the upgrade.

handlers/hop_by_hop.go
domdom82: Sure. Originally,  I copied this code 1:1 from RequestHandler as a pure refactor. But such small optimizations should be fine if semantically equal.

integration/gdpr_test.go
domdom82: that's kind of confusing I know. so here is why this needs to look this it does now:
- The original test looked for the `stdout` log of RequestHandler, which is a json-structured log that prints
```
"RemoteAddr":"-"
```
as a key-value pair.
- With the removal of RequestHandler, all WebSocket requests are now handled via ProxyRoundTripper and httputils.ReverseProxy, as such their output goes to the regular access log - which is unstructured (no json)
- Now, there is no more key "RemoteAddr" but only the order in which [AccessLogRecord](https://github.com/cloudfoundry/gorouter/blob/main/accesslog/schema/access_log_record.go#L251) prints the values to the log
- This happens to be in the order of `<user-agent> <remote addr>` so, user-agent is logged first, then remote addr.
- If remote-addr shall not be logged due to GDPR, we will now have to check for an `"-"` after the user-agent string.
- I therefore had to add a user-agent string to the test so that I would have a string besides `"-"` to look for.

integration/tls_to_backends_test.go
domdom82: Done.

integration/gdpr_test.go
geofffranks: 👍 thanks!

handlers/hop_by_hop.go
domdom82: Done.

domdom82:
@geofffranks @ameowlia any updates / blockers?

domdom82:
hi @ameowlia @geofffranks I've just stumbled across [this code piece in routeservice.go](https://github.com/cloudfoundry/gorouter/blob/main/handlers/routeservice.go#L81-L91) because a customer was complaining they need to rate-limit their WS app using a route-service which is not supported atm.
Do you recall why we don't support WebSockets over route-services?
I didn't double-check but it may be to do with how we implemented WebSockets prior to this PR - so it may be another benefit if we were able to remove that limitation as a follow-up.

geofffranks (approved):

PR #405, opened 2024-04-03
revert: remove deprecated buildnametocertificate call

maxmoehl:
The refactor had unintended side-effects: when selecting a certificate the NameToCertificate map which is built using this call ensured that more specific certificates are preferred over wildcard matches. We rely on that behaviour to present the correct certificate to clients.

This reverts commit 6c6bd520ad939910fe8c0830558f6dd179e7d113.

Resolves: cloudfoundry/routing-release#401

* [ ] I have [run all the unit tests](https://github.com/cloudfoundry/routing-release#running-unit-and-integration-tests).


peanball (approved):
@geofffranks, a small OK from you would be welcome.

PR #408, opened 2024-04-12
Add test to validate cert name selection logic

geofffranks:

Summary
---------------
Adds a test to ensure we don't break the [certificate selection logic that broke](https://github.com/cloudfoundry/routing-release/issues/401) when we removed the BuildNameToCertificate() call in [v0.293.0](https://github.com/cloudfoundry/routing-release/releases/tag/v0.293.0) and [v0.294.0](https://github.com/cloudfoundry/routing-release/releases/tag/v0.294.0). 

Backward Compatibility
---------------
Breaking Change? No

### DIFF
--- a/router/router_test.go
+++ b/router/router_test.go
@@ -2240,6 +2240,34 @@ var _ = Describe("Router", func() {
 			})
 		})
 	})
+
+	FDescribe("cert selection", func() {
+		Context("when multiple overlapping certs are provided to the router", func() {
+			var certPool *x509.CertPool
+			BeforeEach(func() {
+				certPool = x509.NewCertPool()
+				wildcardCert := test_util.CreateCert(fmt.Sprintf("*.%s", test_util.LocalhostDNS))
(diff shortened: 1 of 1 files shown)


geofffranks:
@maxmoehl when you get a chance can you review this + validate it catches the case we regressed on?

maxmoehl (approved):
Yes, this should ensure we don't see this issue again. Thank you very much!

router/router_test.go
```diff
@@ -2240,6 +2240,34 @@ var _ = Describe("Router", func() {
 			})
 		})
 	})
+
+	FDescribe("cert selection", func() {
```
winkingturtle-vmw, line 2244: I think the "Focus" has be removed

maxmoehl: Uhh sorry, didn't notice that. I'll fix it :)

maxmoehl: see #409

PR #410, opened 2024-04-17
feat: tracing

maxmoehl:

Summary
---------------

From time to time we have to investigate situations in which customers complain about individual requests taking a long time (up to 100ms) to pass through gorouter. While investigating ways to find out what causes such delays we came across the built-in tracing functionality of GoLang. Unfortunately the tracing is relatively useless if you don't instruct your code to record certain trace points.

This PR proposes to add an initial set of trace points. Since we don't really know what we are looking for the trace points are a bit scattered around. Please let me know if you think that certain trace points are pointless or if there's any other sections that should have tracing.

Since we have benchmarks in the registry I checked what the impact of the tracing points would be:

| Benchmark | Impact (relative increase in avg of ns/op) |
|---|---|
|BenchmarkRegisterWith100KRoutes-12 | +19% |
|BenchmarkRegisterWithOneRoute-12 | +15% |
|BenchmarkRegisterWithConcurrentLookupWith100kRoutes-12 | +3% |

Note: tracing is *not* enabled, this is just due to the trace points existing and being part of the execution path, they are not recording data.

Given that `BenchmarkRegisterWithConcurrentLookupWith100kRoutes-12` is the closest to what gorouter actually does and it only increases the time per register by ~3% I don't expect any major impact by this change.

Once we have the opportunity to collect some traces on our productive systems we will have more insights into which regions should be traced and which ones are less interesting. We should then re-visit the topic and adjust the trace points accordingly.

Backward Compatibility
---------------

Breaking Change? **No**

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -1,0 +1,42 @@
+# Troubleshooting gorouter
+
+## Profiling and Tracing
+
+A pprof endpoint is exposed via the [debugserver](https://github.com/cloudfoundry/debugserver)
+package and can be configured using the `debug_addr` property. The endpoint allows operators to
+enable profiling or tracing and extract data without changing gorouter itself. The
+[net/http/pprof](https://pkg.go.dev/net/http/pprof#hdr-Usage_examples) package contains examples on
+how to collect the various profiling / tracing data.
+
+After collecting data it can be visualized using the `go` CLI. For best results you should copy the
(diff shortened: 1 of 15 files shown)


maxmoehl:
I will keep the branch around but given the efforts to move towards OTel I don't think it's a good idea to introduce go specific tracing at this point.

PR #411, opened 2024-04-19
feat: make syslog logging useable

maxmoehl:

Summary
---------------

This PR corrects and improves the direct-to-syslog logging of access logs for Gorouter.

* does not log an empty new line when writing to syslog
* allows the operator to configure truncation on syslog, as syslog servers usually have a message limit
* significant performance improvement to the syslog implementation

The syslog implementation in Go has not been maintained by the Go authors since around 2015. While there are forks of the syslog code, all of them are also dormant. This PR inlines and significantly rewrites the syslog implementation, focusing on the following performance improvements:

* No use of the expensive fmt.Sprintf
* Determine all "static" fields (e.g. the PID) on initialization, not per message
* Correct handling of UTF-8 BOM according to [RFC5424](https://datatracker.ietf.org/doc/html/rfc5424)

Limitations:
* Only TCP and UDP are supported, as described in the routing-release property for the syslog network. All other means (unix sockets, etc.) are not supported.

Backward Compatibility
---------------
Breaking Change? **Yes**
 * This PR moves the syslog implementation from the BSD-style syslog format, which is not standardised, to the [RFC5424](https://datatracker.ietf.org/doc/html/rfc5424) format.

### DIFF
--- a/accesslog/syslog/syslog.go
+++ b/accesslog/syslog/syslog.go
@@ -1,0 +1,257 @@
+// This file is part of gorouter of Cloud Foundry. The implementation is a modified version of the
+// Go standard library implementation at log/syslog/syslog.go. Any modifications are licensed under
+// the license of gorouter which can be found in the LICENSE file.
+//
+// Original License:
+//
+// Copyright 2009 The Go Authors.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//    * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//    * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//    * Neither the name of Google LLC nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+// Package syslog implements a syslog writer over UDP and TCP following RFC5424, RFC5426 and
+// RFC6587. It is designed to serve as an access log writer for gorouter and is therefore not
+// general purpose.
+package syslog
+
+import (
+	"bytes"
+	"fmt"
+	"net"
+	"os"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+)
+
+// ByteOrderMark as required by RFC5424
+const ByteOrderMark = "\ufeff"
+
+// The Priority is a combination of the syslog facility and
+// severity. For example, [SeverityAlert] | [FacilityFtp] sends an alert severity
+// message from the FTP facility. The default severity is [SeverityEmerg];
+// the default facility is [FacilityKern].
+type Priority int
+
+const severityMask = 0x07
+const facilityMask = 0xf8
+
+const (
+	// Severity.
+
+	// From /usr/include/sys/syslog.h.
+	// These are the same on Linux, BSD, and OS X.
+	SeverityEmerg Priority = iota
+	SeverityAlert
+	SeverityCrit
+	SeverityErr
+	SeverityWarning
+	SeverityNotice
+	SeverityInfo
+	SeverityDebug
+)
+
+const (
+	// Facility.
+
+	// From /usr/include/sys/syslog.h.
+	// These are the same up to LOG_FTP on Linux, BSD, and OS X.
+	FacilityKern Priority = iota << 3
+	FacilityUser
+	FacilityMail
+	FacilityDaemon
+	FacilityAuth
+	FacilitySyslog
+	FacilityLpr
+	FacilityNews
+	FacilityUucp
+	FacilityCron
+	FacilityAuthPriv
+	FacilityFtp
+	_ // unused
+	_ // unused
+	_ // unused
+	_ // unused
+	FacilityLocal0
+	FacilityLocal1
+	FacilityLocal2
+	FacilityLocal3
+	FacilityLocal4
+	FacilityLocal5
+	FacilityLocal6
+	FacilityLocal7
+)
+
+var (
+	ErrInvalidNetwork  = fmt.Errorf("syslog: invalid network")
+	ErrInvalidPriority = fmt.Errorf("syslog: invalid priority")
+)
+
+// A Writer is a connection to a syslog server.
+type Writer struct {
+	priority string
+	hostname string
+	procid   string
+	appName  string
+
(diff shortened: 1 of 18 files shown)


maxmoehl:
I forgot to update the breaking change indicator after pushing my latest change. This is technically breaking because we move from the BSD-style syslog format which is not standardised to the [RFC5424](https://datatracker.ietf.org/doc/html/rfc5424) format. I'm not sure how man people actually use the syslog implementation today, over UDP each log line also inserts an empty log line due to the (previously dedicated) call to `Write`. Over TCP it did probably work, however, it caused gorouter to perform worse as the old implementation did unnecessary syscalls for each log line written and the formatting was done using `fmt`.

accesslog/syslog/syslog.go
```diff
@@ -0,0 +1,259 @@
...
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+//go:build !windows && !plan9
```
peanball, line 35: because this is now only supporting tcp and udp, do we still have this build limitation?

maxmoehl: I don't think so, I'll drop the constraint.

ameowlia:
Discussion in slack here: https://cloudfoundry.slack.com/archives/C02HNDJB31R/p1732140514501099

maxmoehl:
The old format looks something like this:
```
<34>Oct 11 22:14:15 mymachine MESSAGE
```
and the new format is something like this:
```
<34>1 2003-10-11T22:14:15.003Z mymachine.example.com su - ID47 - \ufeffMESSAGE
```

maxmoehl:
For testing use this config:
```
logging.syslog_addr: 127.0.0.1:514
logging.syslog_network: udp
enable_access_log_streaming: true
```

ameowlia (changes requested):
Before example wihtout deploying any code from when I ran `cf apps` (note the extra line): 
```
2024-12-03T17:21:35Z 7c816b03-30fe-48fc-8112-024f1b7c875d vcap.gorouter[7]: api.z85e8a830.shepherd.lease - [2024-12-03T17:21:35.246128238Z] "GET /v3/apps?order_by=name&per_page=5000&space_guids=f091f24f-9f20-495c-93c0-171b2d5c405a HTTP/1.1" 200 0 424 "-" "cf/8.8.3+3cd802e.2024-10-29 (go1.23.1; amd64 linux)" "34.121.99.84:35278" "10.244.0.133:9024" x_forwarded_for:"34.121.99.84" x_forwarded_proto:"https" vcap_request_id:"41376a16-828e-47b5-6c49-4e9adf5d96aa" response_time:0.032040 gorouter_time:0.000163 app_id:"-" app_index:"-" instance_id:"5a0886b8-5a06-4fda-66ec-db02512ba331" x_cf_routererror:"-" x_b3_traceid:"41376a16828e47b56c494e9adf5d96aa" x_b3_spanid:"6c494e9adf5d96aa" x_b3_parentspanid:"-" b3:"41376a16828e47b56c494e9adf5d96aa-6c494e9adf5d96aa"
2024-12-03T17:21:35Z 7c816b03-30fe-48fc-8112-024f1b7c875d vcap.gorouter[7]:
```

After example (note no extra line)
```
2024-12-03T18:24:08Z 7c816b03-30fe-48fc-8112-024f1b7c875d vcap.gorouter[7] api.z85e8a830.shepherd.lease - [2024-12-03T18:24:08.892876502Z] "GET /v3/apps?order_by=name&per_page=5000&space_guids=f091f24f-9f20-495c-93c0-171b2d5c405a HTTP/1.1" 200 0 424 "-" "cf/8.8.3+3cd802e.2024-10-29 (go1.23.1; amd64 linux)" "34.121.99.84:52634" "10.244.0.133:9024" x_forwarded_for:"34.121.99.84" x_forwarded_proto:"https" vcap_request_id:"83e5beb2-4e2f-430d-7412-b01fa140c0a1" response_time:0.032635 gorouter_time:0.000214 app_id:"-" app_index:"-" instance_id:"8bd9eb77-02e5-4cac-5e12-aa8dea25e775" x_cf_routererror:"-" x_b3_traceid:"83e5beb24e2f430d7412b01fa140c0a1" x_b3_spanid:"7412b01fa140c0a1" x_b3_parentspanid:"-" b3:"83e5beb24e2f430d7412b01fa140c0a1-7412b01fa140c0a1"
```

###  ✅  1. No more blank lines
Great!

### ❓ 2. No difference in format 
I don't see any difference between the before and after logs. Do you know why this is? 

### ⛔ 3. Static Check Failure
When I run `./scripts/test-in-docker.bash` I get the following failure...
```
accesslog/file_and_loggregator_access_logger.go:77:22: io.Writer.Write must not modify the provided buffer, not even temporarily (SA1023)
```

### ✏️  4. Integration test
Is it possible to add an integration test to test this? That way if breaks we will find out in less than 6 years unlike last time.

maxmoehl:
> ❓ 2. No difference in format
> 
> I don't see any difference between the before and after logs. Do you know why this is?

Where did you get those logs from? If, for example, they have passed through another syslog server it could very well be that the syslog server "unifies" the format and you won't be able to tell the difference.

> ⛔ 3. Static Check Failure
> 
> When I run ./scripts/test-in-docker.bash I get the following failure...

Interesting, since this is the file writer I will just change it to be two writes, the original message and a second one for the new line.

ameowlia:
> Where did you get those logs from? If, for example, they have passed through another syslog server it could very well be that the syslog server "unifies" the format and you won't be able to tell the difference.

I turned on UDP for my local rsyslog by un-commenting the UDP stuff in `/etc/rsyslog.conf` on the router VM. Restarted rsyslog (`service rsyslog restart`), then tailed `/var/log` while I ran `cf apps` in another terminal.

peanball:
@ameowlia, yes, this is the format we see on the local to-file output of rsyslogd as well, i.e. it does take the syslog protocol defined fields such as timestamp, source, pid, etc. and re-formats them into the format you are observing when writing to the local file.

This is also expected behavior, as the data sent is a compliant syslog message and is processed by the rsyslogd server, which in turn persists it in the system's regular syslog file as well.

What we've shown with the format written to the wire towards the syslog server has changed from bsd-style (de-facto standard, not an RFC) to [RFC5424](https://datatracker.ietf.org/doc/html/rfc5424). rsyslogd processes either of those and some other formats correctly. So ultimately it's good that the lines you observe in /var/log/syslog did not change. This means, rsyslogd can process the data we send it correctly.

maxmoehl:
@ameowlia I've added an integration test.

integration/access_log_test.go
```diff
@@ -0,0 +1,155 @@
...
+				Expect(res.StatusCode).To(Equal(http.StatusNotFound))
+
+				log := <-logs
+
+				Expect(log).To(ContainSubstring(`x_cf_routererror:"unknown_route"`))
+				Expect(log).To(ContainSubstring(`"GET / HTTP/1.1" 404`))
+				Expect(log).To(ContainSubstring("foobar.cloudfoundry.org"))
+
+				// ensure we don't see any excess access logs
+				Expect(len(logs)).To(Equal(0))
```
ameowlia, line 65: Change "Expect" to "Consistently"

integration/access_log_test.go
```diff
@@ -0,0 +1,155 @@
...
```
ameowlia, line 93: Change "Expect" to "Consistently"

integration/common_integration_test.go
```diff
@@ -176,7 +176,7 @@ func (s *testState) newPostRequest(url string, body io.Reader) *http.Request {
 	return req
 }
 
-func (s *testState) newRequest(url string) *http.Request {
+func (s *testState) newGetRequest(url string) *http.Request {
```
ameowlia, line 179: Can you rebase your changes? This refactor isn't compatible with the latest commits.

ameowlia (changes requested):
Only a few small changes to request!

Changes...
1. Change a few Expects in the tests to Consistently
2. Rebase off latest main branch and continue the rename of newRequest -> newGetRequest

Additional Tests that I did
✅ Integration tests passed
✅ linter passed (when I manually updated the newRequest -> newGetRequest)
✅ Tested the logging.syslog_message_limit property because I forgot to test that last time.

@maxmoehl Ping me on slack when you make these changes so I can merge the changes asap because more incompatible commits come in.

ameowlia (approved):

ameowlia:
dang. Reverting due panics and to races detected : https://ci.funtime.lol/teams/wg-arp-networking/pipelines/routing-release/jobs/unit-and-integration-tests/builds/370

PR #416, opened 2024-05-14
Add test for checking 100 continues + non-200 response codes

geofffranks:
- [ ] Read the [Contributing document](../blob/-/.github/CONTRIBUTING.md).

Summary
---------------
Adds tests to catch the issue seen in https://github.com/cloudfoundry/routing-release/issues/409

Backward Compatibility
---------------
Breaking Change? **Yes/No**
Nope, just adding a test

### DIFF
--- a/proxy/proxy_test.go
+++ b/proxy/proxy_test.go
@@ -2012,7 +2012,7 @@ var _ = Describe("Proxy", func() {
 					conn.WriteResponse(expectRsp)
 
 					rsp := test_util.NewResponse(200)
-					rsp.Body = io.NopCloser(strings.NewReader("DEFGHI"))
+					rsp.Body = io.NopCloser(strings.NewReader("valid-but-unimportant-response-data"))
 					conn.WriteResponse(rsp)
 
 					conn.Close()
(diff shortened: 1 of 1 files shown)


proxy/proxy_test.go
```diff
@@ -2044,14 +2044,70 @@ var _ = Describe("Proxy", func() {
...
+				//since the building of the log record happens throughout the life of the request
+				b, err := os.ReadFile(f.Name())
+				Expect(err).NotTo(HaveOccurred())
+				Expect(string(b)).To(HavePrefix("statusTest - ["))
+				Expect(string(b)).To(ContainSubstring(`"POST / HTTP/1.1" 201 4 6 "-"`))
+				Expect(string(b)).To(ContainSubstring(`x_forwarded_for:"127.0.0.1" x_forwarded_proto:"http" vcap_request_id:`))
+				Expect(string(b)).To(ContainSubstring(`response_time:`))
+				Expect(string(b)).To(ContainSubstring(`app_id:"456"`))
+				Expect(string(b)).To(ContainSubstring(`app_index:"2"`))
+				Expect(b[len(b)-1]).To(Equal(byte('\n')))
```
jrussett, line 2108: Is this line
```
Expect(b[len(b)-1]).To(Equal(byte('\n')))
```
essentially
```
Expect(b).To(HaveSuffix("\n"))
```

Could you please change line 2053 as well then?

jrussett (changes requested):

proxy/proxy_test.go
```diff
@@ -2044,14 +2044,70 @@ var _ = Describe("Proxy", func() {
...
+					Expect(req.ProtoMajor).To(Equal(1))
+					Expect(req.ProtoMinor).To(Equal(1))
+
+					Expect(body).To(Equal("ABCD"))
+
+					expectRsp := test_util.NewResponse(100)
+					conn.WriteResponse(expectRsp)
+
+					rsp := test_util.NewResponse(201)
+					rsp.Body = io.NopCloser(strings.NewReader("DEFGHI"))
```
jrussett, line 2070: It looks like we return the value `DEFGHI` but we only check that there are non-400 status codes in the response. Should we change these strings to something obvious like `valid-unimportant-response-data`?

jrussett (approved):

PR #417, opened 2024-05-22
Fix metric counting incorrect `Content-Length` headers

peanball:

Summary
---------------

Go 1.22 introduced stricter checks for `Content-Length` headers. The header, if set, must not be empty.

This new strict check can be worked around with `debug.httplaxcontentlength`, which will allow empty `Content-Length` headers. In preparation for users disabling this debug flag that Gorouter turns on by default, a metric is emitted to collect requests that _would_ break when the workaround is disabled.

The metric introduced in https://github.com/cloudfoundry/gorouter/pull/412 was erroneous, see https://github.com/cloudfoundry/routing-release/issues/411.

Backward Compatibility
---------------
Breaking Change? **Yes**(ish)

The metric that was recently introduced, called `missing_content_length_header` is misleadingly named and renamed in this PR to `empty_content_length_header` instead.

### DIFF
--- a/handlers/reporter.go
+++ b/handlers/reporter.go
@@ -38,7 +38,7 @@ func (rh *reporterHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request, ne
 		return
 	}
 	if !validContentLength(r.Header) {
-		rh.reporter.CaptureMissingContentLengthHeader()
+		rh.reporter.CaptureEmptyContentLengthHeader()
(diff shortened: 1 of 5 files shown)


peanball:
@mariash, @MarcPaquette, maybe you can have a look as well.

geofffranks (approved):

geofffranks (approved):

PR #419, opened 2024-05-24
Remove unused field

mariash:


peanball:
`go vet` was complaining about two fields I think, `mutex` and `sched`? I'll check the other one.

maxmoehl (approved):

PR #421, opened 2024-06-04
fix: always validate route-service signature if present

maxmoehl:

Backward Compatibility
---------------
Breaking Change? **No**


handlers/routeservice.go
```diff
@@ -62,6 +62,33 @@ func (r *RouteService) ServeHTTP(rw http.ResponseWriter, req *http.Request, next
 	routeServiceURL := reqInfo.RoutePool.RouteServiceUrl()
 	if routeServiceURL == "" {
 		// No route service is associated with this request
+		if r.config.StrictSignatureValidation() && len(req.Header.Values(routeservice.HeaderKeySignature)) > 0 {
+			// Someone is still setting the header, possibly trying to impersonate a route service
+			// request. We will validate the request to ensure the header is valid.
+			// TODO: what do we have to provide as app URL?
```
geofffranks, line 68: Should this TODO still be here?

maxmoehl: Nope :smile:

maxmoehl: That field is never read and we should probably remove it, but the TODO should not be in this commit, I'll remove it.

handlers/routeservice.go
```diff
@@ -62,6 +62,33 @@ func (r *RouteService) ServeHTTP(rw http.ResponseWriter, req *http.Request, next
...
+			rreq := newRequestReceivedFromRouteService("", req.Header)
+			_, err := r.config.ValidateRequest(rreq)
+			if err != nil {
+				AddRouterErrorHeader(rw, "invalid_route_service_signature")
+				if errors.Is(err, routeservice.ErrExpired) {
+					r.errorWriter.WriteError(
+						rw,
+						http.StatusGatewayTimeout,
+						fmt.Sprintf("Failed to validate Route Service Signature: %s", err.Error()),
```
plowin, line 77: For both error messages, maybe we can add the target URL? Sth like this (untested and did not check Sprintf syntax):
```suggestion
						fmt.Sprintf("Failed to validate Route Service Signature to %s: %s", req.URL.String(), err.Error()),
```

It helps finding who is running into this without correlating trace-IDs.

maxmoehl: This string is also what is written in the response that is sent to the client. Given that the access log can be correlated via some ID I'd rather not make the string more verbose.

maxmoehl:
It seems like I've broken one of the other tests, I'm looking into that.

handlers/routeservice.go
plowin: Shouldn't we split the client response and our error log then? I cannot judge the effort and just know the pain when you want to look at multiple data-points but can only get data via the correlation. I am fine keeping it as is, if the effort it too high.

maxmoehl: I think this is more of a general issue with the error reporting around route-services. I will not address it with this pull request as it seems complex, but we should look into this as I also feel the pain of correlating between requests, route-services and error logs.

maxmoehl:
Ignoring the flakiness of my test setup it seems like all relevant tests pass now. @geofffranks please do re-run the suite on your side to be safe 😅

handlers/routeservice.go
b1tamara: @plowin I think the x_cf_routererror in gorouter access_log is very helpful:
`x_cf_routererror:"invalid_route_service_signature"`. The only thing we cannot distinguish between expired signature and invalid in access log.

b1tamara: We can introduce two different `x_cf_routererror`.
- for expired: `x_cf_routererror:"expired_route_service_signature"`
- for other errors: `x_cf_routererror:"invalid_route_service_signature"`

maxmoehl: I've included that.

plowin: Thx for the hints and clarifying! Then I am totally fine and was still stuck to the very initial version, resolving.

proxy/route_service_test.go
```diff
@@ -136,6 +137,33 @@ var _ = Describe("Route Services", func() {
...
+
+			conn := dialProxy(proxyServer)
+
+			req := test_util.NewRequest("GET", "my_host.com", "/", nil)
+			req.Header.Add(routeservice.HeaderKeySignature, "invalid")
+
+			conn.WriteRequest(req)
+
+			res, body := conn.ReadResponse()
+			Expect(res.StatusCode).To(Equal(http.StatusBadRequest))
```
b1tamara, line 162: You can include the check for the Header "X-Cf-RouterError" like:
```Expect(resp.Header().Get("X-Cf-RouterError")).To(Equal(`invalid_route_service_signature`))```

plowin:
Corresponding spec-change: https://github.com/cloudfoundry/routing-release/pull/413

geofffranks (approved):

PR #424, opened 2024-06-14
fix: preserve x-forwarded-host for expect-100-continue req

maxmoehl:

Summary
---------------
The new code path for `Expect : 100-continue` requests causes the X-Forwarded-Host header to be overwritten to the value of the Host header.

This commit will also copy the X-Forwarded-Host from the incoming to the outgoing request to align the behaviour with non-expect-100-continue requests.

Resolves: cloudfoundry/routing-release#420

Backward Compatibility
---------------
Breaking Change? **No**


geofffranks:
Can you add a unit test for this regression so we can avoid breaking this in the future?

maxmoehl:
Will do, can you take a look at the question I raised in the issue?

geofffranks (approved):

PR #428, opened 2024-07-11
Generic Per-Route Options incl. Load Balancing Algorithm

Dariquest:
Summary
---------------
Detailed specification including that of for all the involved components is provided in the [**RFC**](https://github.com/cloudfoundry/community/blob/main/toc/rfc/rfc-0027-generic-per-route-features.md)

Route definition is extended by generic options map with a load balancing algorithm. Options can be optionally specified in the following way with a json key **lb_algo** for the load balancing algorithm, e.g.
```
{
  "host": "127.0.0.1",
  "port": 4567,
  "protocol": "http1",
  "uris": ["url.localhost.routing.cf-app.com"],
  "options": {
    "lb_algo": "least-connection"
  },
  "app": "some_app_guid",
  "stale_threshold_in_seconds": 120,
  "server_cert_domain_san": "some_subject_alternative_name"
}
```

Supported load-balancing algorithms are specified in the Gorouter configuration in the **LoadBalancingStrategies** list. If a load balancing algorithm, provided for a route, is not in this configured list, the pool load balancing algorithm is kept and the error *invalid-endpoint-load-balancing-algorithm-provided-keeping-pool-lb-algo* is logged.

*RegistryMessage* structure is extended by an **Options** field of type *RegistryMessageOpts* structure, currently only containing the LoadBalancingAlgorithm field. *Endpoint* structure is extended by **LoadBalancingAlgorithm** field. 

When initialising new endpoints, the load balancing algorithm is initialised to the value from the corresponding registry message option. When endpoints are registered and added to the pool, the load balancing algorithm of a pool is set to that one of a new endpoint.

*RouteRegistry* structure is extended by a new field **DefaultLoadBalancingAlgorithm**, which is initialised by the Gorouter default load balancing configuration property **LoadBalance**. This default value is a default load balancing algorithm for a new pool.

Since the route options are optional, not providing them has no effect on the existing default load balancing algorithm logic.

Backward Compatibility
---------------
Breaking Change? **No**

Check list
---------------
- [x] Read the [Contributing document](../blob/-/.github/CONTRIBUTING.md)
- [x] I have viewed signed and have submitted the Contributor License Agreement
- [x] I have made this pull request to the main branch
- [x] I have [run all the unit tests](https://github.com/cloudfoundry/routing-release#running-unit-and-integration-tests).

Resolves: cloudfoundry/routing-release#421

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -18,6 +18,7 @@ import (
 	"gopkg.in/yaml.v2"
 
 	"code.cloudfoundry.org/localip"
+	"slices"
 )
 
 const (
@@ -558,6 +559,10 @@ func DefaultConfig() (*Config, error) {
 	return &c, nil
 }
 
+func IsLoadBalancingAlgorithmValid(lbAlgo string) bool {
+	return slices.Contains(LoadBalancingStrategies, lbAlgo)
+}
+
 func (c *Config) Process() error {
 	if c.GoMaxProcs == -1 {
 		c.GoMaxProcs = runtime.NumCPU()
@@ -727,14 +732,7 @@ func (c *Config) Process() error {
 	}
 
 	// check if valid load balancing strategy
-	validLb := false
-	for _, lb := range LoadBalancingStrategies {
-		if c.LoadBalance == lb {
-			validLb = true
-			break
-		}
-	}
-	if !validLb {
+	if !IsLoadBalancingAlgorithmValid(c.LoadBalance) {
 		errMsg := fmt.Sprintf("Invalid load balancing algorithm %s. Allowed values are %s", c.LoadBalance, LoadBalancingStrategies)
 		//lint:ignore SA1006 - for consistency sake
 		return fmt.Errorf(errMsg)
--- a/config/config.go
+++ b/config/config.go
@@ -559,6 +559,7 @@ func DefaultConfig() (*Config, error) {
 	return &c, nil
 }
 
+// IsLoadBalancingAlgorithmValid Check validity of the provided load balancing algorithm
 func IsLoadBalancingAlgorithmValid(lbAlgo string) bool {
 	return slices.Contains(LoadBalancingStrategies, lbAlgo)
 }
--- a/config/config.go
+++ b/config/config.go
@@ -559,7 +559,6 @@ func DefaultConfig() (*Config, error) {
 	return &c, nil
 }
 
-// IsLoadBalancingAlgorithmValid Check validity of the provided load balancing algorithm
 func IsLoadBalancingAlgorithmValid(lbAlgo string) bool {
 	return slices.Contains(LoadBalancingStrategies, lbAlgo)
 }
--- a/config/config.go
+++ b/config/config.go
@@ -732,7 +732,6 @@ func (c *Config) Process() error {
 		c.RouteServiceEnabled = true
 	}
 
-	// check if valid load balancing strategy
 	if !IsLoadBalancingAlgorithmValid(c.LoadBalance) {
 		errMsg := fmt.Sprintf("Invalid load balancing algorithm %s. Allowed values are %s", c.LoadBalance, LoadBalancingStrategies)
 		//lint:ignore SA1006 - for consistency sake
--- a/mbus/subscriber_test.go
+++ b/mbus/subscriber_test.go
@@ -477,6 +477,101 @@ var _ = Describe("Subscriber", func() {
 			Expect(originalEndpoint).To(Equal(expectedEndpoint))
 		})
 
+		Context("when the message contains load balancing algorithm option", func() {
+			JustBeforeEach(func() {
+				sub = mbus.NewSubscriber(natsClient, registry, cfg, reconnected, l)
+				process = ifrit.Invoke(sub)
+				Eventually(process.Ready()).Should(BeClosed())
+			})
+
+			It("endpoint is constructed with the correct load balancing algorithm", func() {
+				var expectedLBAlgo = "round-robin"
+				var msg = mbus.RegistryMessage{
+					Host:     "host",
+					App:      "app",
+					Protocol: "http2",
+					Uris:     []route.Uri{"test.example.com"},
+					Options:  mbus.RegistryMessageOpts{LoadBalancingAlgorithm: expectedLBAlgo},
+				}
+				data, err := json.Marshal(msg)
+				Expect(err).NotTo(HaveOccurred())
+
+				err = natsClient.Publish("router.register", data)
+				Expect(err).ToNot(HaveOccurred())
+
+				Eventually(registry.RegisterCallCount).Should(Equal(2))
+				_, originalEndpoint := registry.RegisterArgsForCall(0)
+				expectedEndpoint := route.NewEndpoint(&route.EndpointOpts{
+					Host:                   "host",
+					AppId:                  "app",
+					Protocol:               "http2",
+					LoadBalancingAlgorithm: expectedLBAlgo,
+				})
+
+				Expect(originalEndpoint).To(Equal(expectedEndpoint))
+			})
+		})
+
+		Context("when the message contains an empty load balancing algorithm option", func() {
+			JustBeforeEach(func() {
+				sub = mbus.NewSubscriber(natsClient, registry, cfg, reconnected, l)
+				process = ifrit.Invoke(sub)
+				Eventually(process.Ready()).Should(BeClosed())
+			})
+
+			It("endpoint is constructed with the empty string load balancing algorithm", func() {
+				var msg = mbus.RegistryMessage{
+					Host:     "host",
+					App:      "app",
+					Protocol: "http2",
+					Uris:     []route.Uri{"test.example.com"},
+					Options:  mbus.RegistryMessageOpts{LoadBalancingAlgorithm: ""},
+				}
+				data, err := json.Marshal(msg)
+				Expect(err).NotTo(HaveOccurred())
+
+				err = natsClient.Publish("router.register", data)
+				Expect(err).ToNot(HaveOccurred())
+
+				Eventually(registry.RegisterCallCount).Should(Equal(2))
+				_, originalEndpoint := registry.RegisterArgsForCall(0)
+				expectedEndpoint := route.NewEndpoint(&route.EndpointOpts{
+					Host:                   "host",
+					AppId:                  "app",
+					Protocol:               "http2",
+					LoadBalancingAlgorithm: "",
+				})
+
+				Expect(originalEndpoint).To(Equal(expectedEndpoint))
+			})
+
+			It("endpoint is constructed with an empty options struct", func() {
+				var msg = mbus.RegistryMessage{
+					Host:     "host",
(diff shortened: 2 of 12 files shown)


registry/registry.go
```diff
@@ -123,6 +125,23 @@ func (r *RouteRegistry) register(uri route.Uri, endpoint *route.Endpoint) route.
 	}
 
 	endpointAdded := pool.Put(endpoint)
+	//Check endpoint for a load balancing algorithm. If it does exist & differs from that of the pool, then take the lb algorithm of the endpoint, otherwise keep the pools lb algorithm.
+	if len(endpoint.LoadBalancingAlgorithm) > 0 && strings.Compare(endpoint.LoadBalancingAlgorithm, pool.LBAlgorithm) != 0 {
```
maxmoehl, line 129: From the [strings.Compare](https://pkg.go.dev/strings#Compare) documentation:

> Compare is included only for symmetry with package bytes. It is usually clearer and **always faster** to use the built-in string comparison operators ==, <, >, and so on.

(emphasis mine)

route/pool.go
```diff
@@ -156,9 +157,10 @@ type EndpointPool struct {
 	NextIdx            int
 	maxConnsPerBackend int64
 
-	random    *rand.Rand
-	logger    logger.Logger
-	updatedAt time.Time
+	random      *rand.Rand
+	logger      logger.Logger
+	updatedAt   time.Time
+	LBAlgorithm string
```
maxmoehl, line 163: Can we please use consistent naming? The property is spelled out at all other places.

registry/registry.go
```diff
@@ -123,6 +125,23 @@ func (r *RouteRegistry) register(uri route.Uri, endpoint *route.Endpoint) route.
...
+		if config.IsLoadBalancingAlgorithmValid(endpoint.LoadBalancingAlgorithm) {
+			pool.Lock()
```
maxmoehl, line 131: The locking needs some improvement, in the previous if statement you already read from the pool without holding a lock.

registry/registry.go
```diff
@@ -123,6 +125,23 @@ func (r *RouteRegistry) register(uri route.Uri, endpoint *route.Endpoint) route.
 	}
 
 	endpointAdded := pool.Put(endpoint)
+	//Check endpoint for a load balancing algorithm. If it does exist & differs from that of the pool, then take the lb algorithm of the endpoint, otherwise keep the pools lb algorithm.
```
maxmoehl, line 128: Comments should be folded at a reasonable length and always start with a space.

config/config.go
```diff
@@ -558,6 +559,11 @@ func DefaultConfig() (*Config, error) {
 	return &c, nil
 }
 
+// IsLoadBalancingAlgorithmValid Check validity of the provided load balancing algorithm
```
maxmoehl, line 562: Please take a look at [Go Doc Comments](https://go.dev/doc/comment) for recommendations on how to format / write them. Given the expressive nature of the function name I would argue that we can leave out the doc here.

config/config.go
```diff
@@ -727,14 +733,7 @@ func (c *Config) Process() error {
 	}
 
 	// check if valid load balancing strategy
```
maxmoehl, line 735: Please delete the comment as well.

maxmoehl (commented):
I've focused on the actual code, I'll review the tests once we have the first round of reviews in.

config/config.go
Dariquest: Thanks, will do.

route/pool.go
Dariquest: I have selected this name to provide consistency to the EndpointPool structure field naming convention, e.g. another property is called maxConnsPerBackend. Naming it LoadBalancingAlgorithm would stand out as the Java styled name in the structure. I see your point though and will consider to rename it.

registry/registry.go
Dariquest: done

registry/registry.go
Dariquest: done

config/config.go
Dariquest: done

Dariquest:
Corresponding issue https://github.com/cloudfoundry/routing-release/issues/421

config/config.go
maxmoehl: I specifically had this quote in mind:

> [...] Go doc comments use complete sentences. For a package comment, that means the [first sentence](https://go.dev/pkg/go/doc/#Package.Synopsis) begins with “Package ”.

In this case, that would translate to something like this:

```suggestion
// IsLoadBalancingAlgorithmValid checks the validity of the provided load balancing algorithm.
```

But the sentence doesn't add any value over the name of the function, so it might as well be left out.

Same goes for:

```go
// OverrulePoolLoadBalancingAlgorithm Overwrites the load balancing algorithm of a pool by that of a specified endpoint, if that is valid.
func (p *EndpointPool) OverrulePoolLoadBalancingAlgorithm(endpoint *Endpoint) {
```

route/pool.go
```diff
@@ -371,11 +377,16 @@ func (p *EndpointPool) removeEndpoint(e *endpointElem) {
 	p.Update()
 }
 
-func (p *EndpointPool) Endpoints(logger logger.Logger, defaultLoadBalance string, initial string, mustBeSticky bool, azPreference string, az string) EndpointIterator {
-	switch defaultLoadBalance {
+func (p *EndpointPool) Endpoints(logger logger.Logger, initial string, mustBeSticky bool, azPreference string, az string) EndpointIterator {
+	switch p.LoadBalancingAlgorithm {
 	case config.LOAD_BALANCE_LC:
+		logger.Debug("endpoint-with-least-connection-lb-algo-added-to-pool")
```
maxmoehl, line 383: This function does not add something to the pool, it returns an iterator for the pool. Please adjust the log messages accordingly.

route/pool.go
```diff
@@ -473,6 +484,25 @@ func (p *EndpointPool) MarshalJSON() ([]byte, error) {
 	return json.Marshal(endpoints)
 }
 
+// OverrulePoolLoadBalancingAlgorithm Overwrites the load balancing algorithm of a pool by that of a specified endpoint, if that is valid.
+func (p *EndpointPool) OverrulePoolLoadBalancingAlgorithm(endpoint *Endpoint) {
+	p.Lock()
+	defer p.Unlock()
+	if len(endpoint.LoadBalancingAlgorithm) > 0 && endpoint.LoadBalancingAlgorithm != p.LoadBalancingAlgorithm {
+		if config.IsLoadBalancingAlgorithmValid(endpoint.LoadBalancingAlgorithm) {
+			// Multiple apps can have the same route, a pool will get the last endpoint's algorithm
```
maxmoehl, line 493: What should this comment tell the reader? The fact that a single route can be added to many apps does not interfere with the per-route options as those are attached to the route, not the app.

mbus/subscriber_test.go
```diff
@@ -477,6 +477,101 @@ var _ = Describe("Subscriber", func() {
...
+			})
+
+			It("endpoint is constructed with the empty string load balancing algorithm", func() {
+				var msg = mbus.RegistryMessage{
+					Host:     "host",
+					App:      "app",
+					Protocol: "http2",
+					Uris:     []route.Uri{"test.example.com"},
+					Options:  mbus.RegistryMessageOpts{LoadBalancingAlgorithm: ""},
+				}
```
maxmoehl, line 529: This and the test below are identical in go. All fields of a struct will have their default values whether they are mentioned or not, since the options struct contains a string field for the load balancing algorithm, that field will be `""` even if you don't specify it explicitly.

maxmoehl: This would be equivalent as well:
```suggestion
				var msg = mbus.RegistryMessage{
					Host:     "host",
					App:      "app",
					Protocol: "http2",
					Uris:     []route.Uri{"test.example.com"},
				}
```

route/pool.go
```diff
@@ -473,6 +484,25 @@ func (p *EndpointPool) MarshalJSON() ([]byte, error) {
 	return json.Marshal(endpoints)
 }
 
+// OverrulePoolLoadBalancingAlgorithm Overwrites the load balancing algorithm of a pool by that of a specified endpoint, if that is valid.
+func (p *EndpointPool) OverrulePoolLoadBalancingAlgorithm(endpoint *Endpoint) {
```
maxmoehl, line 488: Since you only require the `Endpoint.LoadBalancingAlgorithm` field I would only pass in that string and rename the function, maybe `SetLoadBalancingAlgorithm`.

route/pool_test.go
```diff
@@ -231,6 +232,110 @@ var _ = Describe("EndpointPool", func() {
...
+		It("is an empty string and the load balancing algorithm of a pool is kept", func() {
+			expectedLBAlgo := config.LOAD_BALANCE_RR
+			pool := route.NewPool(&route.PoolOpts{
+				Logger:                 logger,
+				LoadBalancingAlgorithm: expectedLBAlgo,
+			})
+			endpoint := route.NewEndpoint(&route.EndpointOpts{
+				Host: "host-1", Port: 1234,
+				RouteServiceUrl:        "url",
+				LoadBalancingAlgorithm: "",
```
maxmoehl, line 303: Same as above, this test and the one below are equivalent.

route/pool.go
Dariquest: done

mbus/subscriber_test.go
Dariquest: done

route/pool_test.go
Dariquest: done

route/pool.go
Dariquest: done

config/config.go
Dariquest: done

registry/registry.go
```diff
@@ -123,6 +125,19 @@ func (r *RouteRegistry) register(uri route.Uri, endpoint *route.Endpoint) route.
...
```
stefanlay: typo: pools - > pool's

stefanlay (changes requested):
Found a small typo, loogs good overall.

Dariquest:
> Found a small typo, looks good overall.

Thanks!

stefanlay (approved):

registry/registry.go
Dariquest: @maxmoehl The locking was intentionally done only for write operations, to avoid unnecessary locks for a frequent operation with only a reading access to the pool. Could you provide an example, where this lock strategy could lead to a problem or an inconsistent state, except that we only overwrite a value, which was currently already changed? The result would be correct in this case though. @peanball For the case, that you a bored in your vacation and also want to discuss :)

peanball: You'll still need a read lock to avoid inconsistent read during write. This can be at field or even word level. 
The write is not guaranteed to be atomic, so reads need to have locks too that are exclusive with write locks (and vice versa).

Dariquest: The correct lock strategy is commited and I am resolving this conversation. Thanks!

PR #429, opened 2024-08-06
add support for tls 1.3 cipher suites in openssl format ... even though they are technically never used.

ameowlia:

Summary
---------------
When tls 1.3 is used, golang (and thus gorouter) will always use the [cipher suites listed here](https://docs.vmware.com/en/VMware-Tanzu-Application-Service/4.0/tas-for-vms/securing-traffic.html#tls-v13-7), no matter what the operator provides [here](https://github.com/cloudfoundry/routing-release/blob/69f705daddb39410ed7465eb62d8e73a57f0676b/jobs/gorouter/spec#L213-L220). So technically the operator doesn’t need to provide any tls 1.3 cipher suites.

However, currently, if an operator _does_ provide a tls 1.3 cipher suite in openssl format, gorouter will fail. This is a bad user experience. 

Even though gorouter will ignore these cipher suites and always use golang's defaults for tls 1.3 we shouldn't fail if they are provided.

Backward Compatibility
---------------
Breaking Change? No


MarcPaquette (approved):

config/config.go
```diff
@@ -835,6 +836,9 @@ func (c *Config) processCipherSuites() ([]uint16, error) {
 		"ECDHE-ECDSA-CHACHA20-POLY1305":          0xcca9,
 		"TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305":   0xcca8,
 		"TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305": 0xcca9,
+		"TLS13-AES-128-GCM-SHA256":               0x1301,
+		"TLS13-AES-256-GCM-SHA384":               0x1302,
+		"TLS13-CHACHA20-POLY1305-SHA256":         0x1303,
```
geofffranks, line 841: Where did you get these names from? I'm seeing something different on https://docs.openssl.org/3.0/man1/openssl-ciphers/#tls-v13-cipher-suites:

```
TLS_AES_128_GCM_SHA256                     TLS_AES_128_GCM_SHA256
TLS_AES_256_GCM_SHA384                     TLS_AES_256_GCM_SHA384
TLS_CHACHA20_POLY1305_SHA256               TLS_CHACHA20_POLY1305_SHA256
TLS_AES_128_CCM_SHA256                     TLS_AES_128_CCM_SHA256
TLS_AES_128_CCM_8_SHA256                   TLS_AES_128_CCM_8_SHA256
```

geofffranks: Given that those names match the IANA names, we probbaly don't need to add them. But it's a little annoying that there seem to be multiple sources of truth for what cipher suite names openssl supports.

ameowlia: I got them from our docs: https://docs.vmware.com/en/VMware-Tanzu-Application-Service/4.0/tas-for-vms/securing-traffic.html#tls-v13-7

ameowlia: ugh looks like our docs are wrong :)

PR #432, opened 2024-08-13
Add delay for gorouter integration test to fix flakes

geofffranks:

Summary
---------------
Waits for echoApp to start up and be ready before moving on to the actual test, to avoid 404s seen in CI, but not local unit tests.

Backward Compatibility
---------------
Breaking Change? No


ebroberson (approved):
LGTM

PR #439, opened 2024-09-17
Set ReadHeaderTimeout on gorouter and route-services servers

mariash:

Summary
---------------
Add option to configure ReadHeaderTimeout on gorouter and route-services server.

Backward Compatibility
---------------
Breaking Change? **No**
Default value is 0, this change is not breaking. Change in the release will be breaking.


geofffranks (changes requested):
one tiny issue, otherwise lookds good

geofffranks (approved):

Issue #437, opened 2024-09-10
Enhance gorouter retry logic


b1tamara:
### Current behavior

- balancing_algorithm: round_robin
- cf app with two instances
- each instance has max_connection limit = 1000 
- app code:
```
import express from 'express';
import http from 'http';
import { setTimeout } from 'timers/promises';

const app = express();

app.get('/', async (req, res) => {
    await setTimeout(5000);
    res.send('Hello, World!');
});

const server = http.createServer(app);
server.maxConnections = 1000;

const PORT = process.env.PORT || 3000;
server.listen(PORT, () => {
    console.log(`Server is running on http://localhost:${PORT}`);
    console.log(`Max connections: ${server.maxConnections}`);
});
```

Start to send requests:
`hey -c 1000 -n 1000 -H "X-Cf-App-Instance":"1edeb42a-076a-41a2-84a6-a70056e9938e:0" "https://retry-test.<domain>"`
After 1-2 seconds:
`hey -c 1000 -n 1000 "https://retry-test.<domain>"`

All 1000 requests with the CF-App-Instance Header are successful. 
Ca. 1/4 of request without  CF-App-Instance Header finished with 502 as gorouter sent retries to the same app instance.

Observations: 
- Every RequestInfo has a pointer to the [EndpointPool](https://github.com/cloudfoundry/gorouter/blob/6331451d30dbf996cc9132ab617f07b25707b06b/handlers/lookup.go#L108). For the same uri(route) it points to the same EndpointPool. 
- The issue is related to the iteration [over pool endpoints](https://github.com/cloudfoundry/gorouter/blob/6331451d30dbf996cc9132ab617f07b25707b06b/route/roundrobin.go#L74). 
- Gorouter gets a lot of concurrent requests to the same route. 
- Max_retries = number of app instances.
- A couple of seconds the app instance with instance_id = 0 cannot accept new connections because of max connection limit in the app code. 

Found logs (written by the same gorouter):
```
"timestamp":"2024-09-10T09:46:47.350182234Z","message":"backend","data":{"instance_id":"0"},"attempt":1,"vcap_request_id":"703d8ce7-b187-41b6-4a02-a122d1ee8b60"

"timestamp":"2024-09-10T09:46:47.350406060Z","message":"backend","data":{"instance_id":"1"},"attempt":1,"vcap_request_id":"ffed058e-a289-4ca9-7d22-0e345c2bd54f"

[...]

PR #443, opened 2024-10-04
fix: Enhance gorouter retry logic (#437)

b1tamara:

Summary
---------------
The fix for the issue described in https://github.com/cloudfoundry/gorouter/issues/437. 
- We propose to use the pool index to "seed" each iterator and next calls on that iterator will then loop over the entire pool exactly once (without to affect the global pool iterator).
- Implement additional tests for two parallel iterator 
- Adapt the test cases in round_robin_test.go which were bound to pool nextIndex strongly (e.g. "RoundRobin Next when locally-optimistic mode on the first attempt when the pool has multiple in the same AZ as the router when all AZ-local endpoints have errors it resets the errors and returns one of the endpoints regardless of AZ"  to check not a specific CanonicalAddr but to validate that one endpoint from other AZ has been returned)

Backward Compatibility
---------------
Breaking Change? **No**

### DIFF
--- a/route/roundrobin.go
+++ b/route/roundrobin.go
@@ -3,7 +3,6 @@ package route
 import (
 	"context"
 	"log/slog"
-	"math/rand"
 	"sync"
 	"time"
 )
@@ -12,7 +11,6 @@ type RoundRobin struct {
 	logger *slog.Logger
 	pool   *EndpointPool
 	lock   *sync.Mutex
-	random *rand.Rand
 
 	initialEndpoint       string
 	mustBeSticky          bool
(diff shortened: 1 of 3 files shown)


route/roundrobin.go
```diff
@@ -3,32 +3,44 @@ package route
...
+
+	nextIdx int
 }
 
 func NewRoundRobin(logger *slog.Logger, p *EndpointPool, initial string, mustBeSticky bool, locallyOptimistic bool, localAvailabilityZone string) EndpointIterator {
 	return &RoundRobin{
 		logger:                logger,
 		pool:                  p,
+		lock:                  &sync.Mutex{},
+		random:                rand.New(rand.NewSource(time.Now().UnixNano())),
```
hoffmaen, line 31: Do we use the `ranomd` field somewhere?

b1tamara: No. You are right, it should be removed.

hoffmaen (approved):

ameowlia:
@b1tamara can you add some comments to your tests about exactly what the different loops are doing? It took me a bit of looking to grok them 😅.

Other than that, it LGTM.

b1tamara:
@ameowlia 
Could you please take a look again? I tryed to add more text context with By() to the new test to make it more readable/clear.

ameowlia (approved):

PR #446, opened 2024-10-22
Restore Docs removed by sync-readme task

a18e:
- [x] Read the [Contributing document](../blob/-/.github/CONTRIBUTING.md).

Summary
---------------

In https://github.com/cloudfoundry/gorouter/pull/423, the `sync-readme` task removed a lot of relevant information from the [old README](https://github.com/cloudfoundry/gorouter/blob/ddb5a02d761fc62457740e67a8abeef3eca9bef0/README.md) that was never moved to the new docs structure.

This PR restores most of the content and puts it into the new structure under `docs/`.

Some caveats:

* I have tried to move over everything apart from a couple of sections that I know are outdated.
* I am happy to change the structure that I have introduced, looking for feedback.
* I cannot guarantee that all restored sections are still valid/up-to-date.
* I currently don't know if I need to manually adjust the root [`README.md`](https://github.com/cloudfoundry/gorouter/blob/5174c184566bddac55a2fa42da4f67088de2c097/README.md?plain=1#L36-L40) (especially the TOC there). It does say that the file is autogenerated, but I can't find any more information on the `sync-readme` task other than [the linked metadata.yml](https://github.com/cloudfoundry/wg-app-platform-runtime-ci/blob/c83c224ad06515ed52f51bdadf6075f56300ec93/shared/tasks/sync-readme/metadata.yml) file.
* I have also removed the Table of Contents (previously autgenerated by `vim-markdown-toc GFM`) since it's not consistently used across the various docs in the org (and I believe it to be redundant to IDE or github.com TOC features). I can restore it or remove it from [02-nats-configuration.md](https://github.com/cloudfoundry/gorouter/blob/5174c184566bddac55a2fa42da4f67088de2c097/docs/02-nats-configurations.md?plain=1#L7-L14) as well, just need to know which is preferred.

Backward Compatibility
---------------
Breaking Change? **No**

PR #447, opened 2024-10-22
Show LB Algorithm in RoutingTable (`/routes` page)

a18e:

Summary
---------------
Adds `load_balancing_algorithm` to `Endpoint` JSON representation to display the LB algorithm in the `/routes` page. This simplifies debugging of related issues.
The option to set route-specific load balancing algorithms was introduced in https://github.com/cloudfoundry/gorouter/pull/428.

Backward Compatibility
---------------
Breaking Change? **No**

(According to the [Gorouter docs](https://github.com/cloudfoundry/gorouter/pull/446/files#diff-6b5dcfee48e174b2716e3ac3adcbc21860aab85bdf03789929f4eb64d2e32f9eR46-R48), the /routes endpoint is an internal feature, so it shouldn't be a breaking change even though we're adding a JSON property.)

### DIFF
--- a/route/pool.go
+++ b/route/pool.go
@@ -517,16 +517,17 @@ func (e *endpointElem) isOverloaded() bool {
 
 func (e *Endpoint) MarshalJSON() ([]byte, error) {
 	var jsonObj struct {
-		Address             string            `json:"address"`
-		AvailabilityZone    string            `json:"availability_zone"`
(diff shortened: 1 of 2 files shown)


MarcPaquette (approved):
LGTM. Thanks for the contribution!

PR #452, opened 2024-10-31
Fix Gorouter's debugserver endpoint to allow log level changes at runtime

MarcPaquette:
We have a debugserver in gorouter + explicit code trying to support reconfiguring the log level on the fly. Being able to do this helps operators + support troubleshoot live issues without having to restart Gorouters. This has been broken since ~2017. 

This change allows CF Administrators to change gorouter's logging on the fly to triage issue in much finer detail.

This is related to the change in the debug server https://github.com/cloudfoundry/debugserver/pull/72

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -92,6 +92,8 @@ SetLoggingLevel dynamically sets the logging level at runtime. See https://githu
 for possible logging levels.
 */
 func SetLoggingLevel(level string) {
+	mutex.Lock()
+	defer mutex.Unlock()
 	zapLevel, err := zapcore.ParseLevel(level)
 	if err != nil {
 		panic(err)


logger/logger.go
```diff
@@ -95,7 +96,27 @@ func SetLoggingLevel(level string) {
...
+func (loggingConf DynamicLoggingConfig) SetMinLevel(level lager.LogLevel) {
+	Conf.level.SetLevel(toZapLevel(level))
+}
+
+func toZapLevel(level lager.LogLevel) zapcore.Level {
+	switch level {
+	case lager.DEBUG:
+		return zapcore.DebugLevel
+	case lager.INFO:
+		return zapcore.InfoLevel
```
andy-a-d-nguyen, line 112: *Small suggestion*

If `zapcore.InfoLevel` is the default case, I don't think you need to add a specific case for `lager.INFO` unless there are explicit reasons for doing so.

MarcPaquette: I thought about it, and chose to go explicit over implicit in the handling of the lager libraries while also being verbose about the log leveling hierarchy and translations. There are only a 4 cases with lager to the 6 with zapcore. The lager interface is the one we need to satisfy, so handling each was the priority in my mind.

["Explicit is better than implicit"](https://peps.python.org/pep-0020/) mentality.

logger/logger.go
```diff
@@ -8,13 +8,14 @@ import (
...
 
+	"code.cloudfoundry.org/lager/v3"
 	"go.uber.org/zap"
 	"go.uber.org/zap/exp/zapslog"
 	"go.uber.org/zap/zapcore"
 )
 
 var (
-	conf        dynamicLoggingConfig
+	Conf        DynamicLoggingConfig
```
ameowlia, line 18: why is this exported now? I assume for the change in main.go? 

exposing the variable means anyone anywhere can change it. what about adding a getter function instead?

ameowlia (changes requested):

logger/logger.go
MarcPaquette: We need the satisfy the debugserver's reconfigurable sink: https://github.com/cloudfoundry/debugserver/blob/main/server.go#L24-L26

The [base gorouter logger is inited at the package level and privately](https://github.com/cloudfoundry/gorouter/blob/main/logger/logger.go#L67-L69) puts us in a bit of an odd spot with further changes to the logger. I exported the config as we need the `debugserver` to actually change the config via `SetMinLevel()`

Basically since the `grlogger.baseLogger` is [initialized on importation via `init()`](https://github.com/cloudfoundry/gorouter/blob/main/logger/logger.go#L67-L69)  and private to the package, I pushed the config out as public as we need something to satisfy the `ReconfigurableSinkInterface`.  A getter doesn't quite fit the pattern. There is an argument for pushing it further down, but this was a bit simpler.

I went with the config as it's the area where adding the method to satisfy the `ReconfigurableSinkInterface` was the least intrusive.

logger/logger.go
```diff
@@ -91,11 +92,33 @@ SetLoggingLevel dynamically sets the logging level at runtime. See https://githu
 for possible logging levels.
 */
 func SetLoggingLevel(level string) {
+	mutex.Lock()
+	defer mutex.Unlock()
```
MarcPaquette, line 96: @ameowlia 
I added the Mutex here since there are a few different ways the logging level can get changed.

MarcPaquette:
@ameowlia Is this okay to merge in now?

PR #453, opened 2024-11-07
Introduce response header limits

maxmoehl:

Summary
---------------
* Fix the header accounting on requests.
* (internal) Rename the config field MaxHeaderBytes to MaxRequestHeaderBytes.
* Add MaxRequestHeaderBytes and the corresponding logic.

A request with too large response headers will now look like this:

```sh
$ curl -i "https://net-tools.cf.local/headers?foobarbazz=$(python3 -c 'print("a" * 1000)')"
HTTP/2 502
content-type: text/plain; charset=utf-8
x-cf-routererror: endpoint_failure (net/http: HTTP/1.x transport connection broken: net/http: server response headers exceeded 1024 bytes; aborted)
x-content-type-options: nosniff
date: Thu, 07 Nov 2024 09:24:09 GMT
content-length: 67

```

Backward Compatibility
---------------
Breaking Change? **No**

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -460,7 +460,7 @@ type Config struct {
 	DisableKeepAlives            bool `yaml:"disable_keep_alives"`
 	MaxIdleConns                 int  `yaml:"max_idle_conns,omitempty"`
 	MaxIdleConnsPerHost          int  `yaml:"max_idle_conns_per_host,omitempty"`
-	MaxHeaderBytes               int  `yaml:"max_header_bytes"`
+	MaxRequestHeaderBytes        int  `yaml:"max_request_header_bytes"`
 	KeepAlive100ContinueRequests bool `yaml:"keep_alive_100_continue_requests"`
 
 	HTTPRewrite HTTPRewrite `yaml:"http_rewrite,omitempty"`
--- a/config/config.go
+++ b/config/config.go
@@ -461,6 +461,7 @@ type Config struct {
 	MaxIdleConns                 int  `yaml:"max_idle_conns,omitempty"`
 	MaxIdleConnsPerHost          int  `yaml:"max_idle_conns_per_host,omitempty"`
 	MaxRequestHeaderBytes        int  `yaml:"max_request_header_bytes"`
+	MaxResponseHeaderBytes       int  `yaml:"max_response_header_bytes"`
 	KeepAlive100ContinueRequests bool `yaml:"keep_alive_100_continue_requests"`
 
 	HTTPRewrite HTTPRewrite `yaml:"http_rewrite,omitempty"`
--- a/config/config.go
+++ b/config/config.go
@@ -462,6 +462,8 @@ type Config struct {
 	MaxIdleConnsPerHost          int  `yaml:"max_idle_conns_per_host,omitempty"`
 	MaxRequestHeaderBytes        int  `yaml:"max_request_header_bytes"`
 	MaxResponseHeaderBytes       int  `yaml:"max_response_header_bytes"`
+	MaxRequestHeaders            int  `yaml:"max_request_headers"`
+	MaxResponseHeaders           int  `yaml:"max_response_headers"`
 	KeepAlive100ContinueRequests bool `yaml:"keep_alive_100_continue_requests"`
 
 	HTTPRewrite HTTPRewrite `yaml:"http_rewrite,omitempty"`
--- a/handlers/max_request_size_test.go
+++ b/handlers/max_request_size_test.go
@@ -166,6 +166,17 @@ var _ = Describe("MaxRequestSize", func() {
 			Expect(result.StatusCode).To(Equal(http.StatusRequestHeaderFieldsTooLarge))
 		})
 	})
+	Context("when a repeated header has a short value and long key taking it over the limit", func() {
+		BeforeEach(func() {
+			for i := 0; i < 10; i++ {
+				header.Add("foobar", "m")
+			}
+		})
+		It("throws an http 431", func() {
+			handleRequest()
+			Expect(result.StatusCode).To(Equal(http.StatusRequestHeaderFieldsTooLarge))
+		})
+	})
 	Context("when enough normally-sized headers put the request over the limit", func() {
 		BeforeEach(func() {
 			header.Add("header1", "smallRequest")
--- a/handlers/max_request_size_test.go
+++ b/handlers/max_request_size_test.go
@@ -63,7 +63,7 @@ var _ = Describe("MaxRequestSize", func() {
 
 	BeforeEach(func() {
 		cfg = &config.Config{
-			MaxHeaderBytes:           89,
+			MaxRequestHeaderBytes:    89,
 			LoadBalance:              config.LOAD_BALANCE_RR,
 			StickySessionCookieNames: config.StringSet{"blarg": struct{}{}},
 		}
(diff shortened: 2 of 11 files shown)


maxmoehl:
This is the first set of changes, next up is limiting the number of headers as outlined in the issue.

maxmoehl:
I'm not sure whether we want to rename the gorouter yaml property from `max_header_bytes` to `max_request_header_bytes` or whether that's already considered breaking?

maxmoehl:
Note to the reviewer: I've split the changes into review-able commits focusing on one aspect each so it's easier to review the individual changes :)

config/config.go
```diff
@@ -460,7 +460,10 @@ type Config struct {
 	DisableKeepAlives            bool `yaml:"disable_keep_alives"`
 	MaxIdleConns                 int  `yaml:"max_idle_conns,omitempty"`
 	MaxIdleConnsPerHost          int  `yaml:"max_idle_conns_per_host,omitempty"`
-	MaxHeaderBytes               int  `yaml:"max_header_bytes"`
+	MaxRequestHeaderBytes        int  `yaml:"max_header_bytes"`
+	MaxResponseHeaderBytes       int  `yaml:"max_response_header_bytes"`
+	MaxRequestHeaders            int  `yaml:"max_request_headers"`
```
peanball, line 465: for consistency's sake, we could call this `max_headers`?

I know the missing `request` in `max_(request_)header_bytes` is for backward compatibility, but having `request` in one property and not the other seems also inconsistent.

proxy/proxy.go
```diff
@@ -84,25 +84,27 @@ func NewProxy(
...
-			ExpectContinueTimeout: 1 * time.Second,
+			DialContext:            dialer.DialContext,
+			DisableKeepAlives:      cfg.DisableKeepAlives,
+			MaxIdleConns:           cfg.MaxIdleConns,
+			IdleConnTimeout:        90 * time.Second, // setting the value to golang default transport
+			MaxIdleConnsPerHost:    cfg.MaxIdleConnsPerHost,
+			DisableCompression:     true,
+			TLSClientConfig:        routeServiceTLSConfig,
+			ExpectContinueTimeout:  1 * time.Second,
+			MaxResponseHeaderBytes: int64(cfg.MaxResponseHeaderBytes),
```
peanball, line 107: just because the ports were recently changed to uint, should we have uint here as well?

proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -391,6 +424,21 @@ func (rt *roundTripper) selectEndpoint(iter route.EndpointIterator, request *htt
 	return endpoint, nil
 }
 
+func checkResponseHeaders(maxCount int, headers http.Header) error {
+	if maxCount > 0 {
+		hdrCount := 0
+		for _, vv := range headers {
+			hdrCount += len(vv)
```
peanball, line 431: a small comment here would be nice: we count all instances of repeated headers, and all comma separated lists. According to the RFC they're equivalent.

This should be equivalent to the comments on the request size, which explain nicely what happens and what we limit.

integration/header_test.go
```diff
@@ -197,4 +198,27 @@ var _ = Describe("Headers", func() {
...
+				testApp.Start()
+				testState.register(testApp.Server, testAppRoute)
+			})
+
+			It("fails with 502 when the app exceeds the limit", func() {
+				req := testState.newRequest(fmt.Sprintf("http://%s", testAppRoute))
+				resp, err := testState.client.Do(req)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(resp.StatusCode).To(Equal(http.StatusBadGateway))
+				Expect(resp.Header).To(HaveKey("X-Cf-Routererror"))
```
peanball, line 220: should we check here also for the error message to make sure the right router error is reported?

handlers/max_request_size_test.go
```diff
@@ -166,6 +167,28 @@ var _ = Describe("MaxRequestSize", func() {
...
+	})
+	Context("when there are too many headers", func() {
+		BeforeEach(func() {
+			for i := 0; i < 16; i++ {
+				header.Add("f", "m")
+			}
+		})
+		It("throws an http 431", func() {
+			handleRequest()
+			Expect(result.StatusCode).To(Equal(http.StatusRequestHeaderFieldsTooLarge))
```
peanball, line 189: do we also have a router error header in the response? if so, can we also add a check for that?

peanball (changes requested):
requesting to review some naming consistency and code comments for clarification (alignment with existing comment density)

config/config.go
maxmoehl: See also my [previous comment](https://github.com/cloudfoundry/gorouter/pull/453#issuecomment-2467674019). If we are fine with adjusting the coding and only guaranteeing the compatibility on release level, that would be my preferred solution.

proxy/proxy.go
maxmoehl: Well the transport is from the go standard library, we can't change the type there. I don't see much value in using a uint over a regular int, you won't be able to set a higher value because the stdlib only accepts signed ints and we just consider negative values as no limit (which means some default is put in place).

proxy/round_tripper/proxy_round_tripper.go
maxmoehl: To make sure we are on the same page, this:

```
Foo: bar
Foo: bar, baz
```

will result in:

```go
map[string][]string{
	"Foo": []string{"bar", "bar, baz"},
}
```

To my understanding this is in conflict with the RFC but there's probably a good reason for that? At any rate: this allows us to deduce the number of header keys which for us is good, otherwise we would have no way of knowing how often a header key occurred.

In any case: here we are just interested in the number of keys which is equal to the number of values in the entire map. I'm adding a comment similar to the one present in the MaxRequestSize middleware.

peanball: you're right. I wrote the thing about the RFC, then read the comment on the other code that explains what's happening and that it's counting the number of values, not the number of "header lines".

Either way, as long as it's consistent it's good.

adding the comment is what I wanted. Thanks!

proxy/proxy.go
peanball: now I see that this is a golang builtin. Disregard my comment.

config/config.go
peanball: This might also be a question for the routing-release, right?

In the end we don't intend to configure gorouter directly but only via the routing release. The conversion / could also be done there, i.e. legacy parsing of `max_header_bytes`, but allow `max_request_header_bytes` also, and map both to gorouter's `max_request_header_bytes`?

maxmoehl: That's probably the best solution, I'll see what I can do.

maxmoehl: See the routing-release PR: https://github.com/cloudfoundry/routing-release/pull/443

handlers/max_request_size_test.go
maxmoehl: yes, I've added it to one of the tests.

peanball (dismissed):

maxmoehl:
Once the routing-release PR is ready as well I will merge both.

peanball (approved):

proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -38,7 +38,10 @@ const (
 	AuthNegotiateHeaderCookieMaxAgeInSeconds = 60
 )
 
-var NoEndpointsAvailable = errors.New("No endpoints available")
+var (
+	NoEndpointsAvailable   = errors.New("No endpoints available")
+	TooManyResponseHeaders = errors.New("too many response headers")
```
ameowlia, line 43:
```suggestion
	TooManyResponseHeaders = errors.New("Too many response headers")
```

ameowlia (approved):

PR #454, opened 2024-11-08
feat: log local TCP address

maxmoehl:
Summary
---------------

When troubleshooting issues related to keepalive it is often useful to be able to correlate multiple requests made over the same connection. A first step was to introduce the information on whether the request was made on a re-used connection or not which was done with a7dbf849. This commit adds the local address of the TCP connection which adds the capability to distinguish between multiple, parallel keep-alive connections by looking at the client port of gorouter for the backend connection.

Backward Compatibility
---------------
Breaking Change? **Yes**

### DIFF
--- a/accesslog/schema/access_log_record.go
+++ b/accesslog/schema/access_log_record.go
@@ -110,6 +110,7 @@ type AccessLogRecord struct {
 	LogAttemptsDetails     bool
 	FailedAttempts         int
 	RoundTripSuccessful    bool
+	ExtraFields            []string
 	record                 []byte
 
 	// See the handlers.RequestInfo struct for details on these timings.
@@ -124,6 +125,8 @@ type AccessLogRecord struct {
 	TlsHandshakeFinishedAt      time.Time
 	AppRequestFinishedAt        time.Time
 	FinishedAt                  time.Time
+
+	LocalAddress string
 }
 
 func (r *AccessLogRecord) formatStartedAt() string {
@@ -317,6 +320,17 @@ func (r *AccessLogRecord) makeRecord(performTruncate bool) []byte {
 		// #nosec  G104 - ignore errors from writing the access log as it will only cause more errors to log this error
 		b.WriteString(`backend_time:`)
 		b.WriteDashOrFloatValue(r.successfulAttemptTime())
+	}
+
+	// We have to consider the impact of iterating over a list. This technically allows to repeat
+	// some of the fields but it allows us to iterate over the list only once instead of once per
+	// field when we perform a [slices.Contains] check.
+	for _, field := range r.ExtraFields {
+		switch field {
+		case "local_address":
+			b.WriteString(`local_address:`)
+			b.WriteDashOrStringValue(r.LocalAddress)
+		}
 	}
 
 	b.AppendSpaces(false)
--- a/accesslog/schema/access_log_record_test.go
+++ b/accesslog/schema/access_log_record_test.go
@@ -314,6 +314,43 @@ var _ = Describe("AccessLogRecord", func() {
 				Eventually(r).Should(Say(`x_cf_routererror:"-"`))
 			})
 		})
+
+		Context("when extra_fields is set to [local_address]", func() {
+			Context("and the local address is empty", func() {
+				It("makes a record with the local address set to -", func() {
+					record.ExtraFields = []string{"local_address"}
+
+					r := BufferReader(bytes.NewBufferString(record.LogMessage()))
+					Eventually(r).Should(Say(`local_address:"-"`))
+				})
+			})
+			Context("and the local address contains an address", func() {
+				It("makes a record with the local address set to that address", func() {
+					record.ExtraFields = []string{"local_address"}
+					record.LocalAddress = "10.0.0.1:34823"
+
+					r := BufferReader(bytes.NewBufferString(record.LogMessage()))
+					Eventually(r).Should(Say(`local_address:"10.0.0.1:34823"`))
+				})
+			})
+		})
+
+		Context("when extra_fields is set to [foobarbazz]", func() {
+			It("ignores it", func() {
+				record.ExtraFields = []string{"foobarbazz"}
+				record.LocalAddress = "10.0.0.1:34823"
+
+				r := BufferReader(bytes.NewBufferString(record.LogMessage()))
+				Consistently(r).ShouldNot(Say("foobarbazz"))
+			})
+			It("does not log local_address", func() {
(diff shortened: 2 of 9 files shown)


maxmoehl:
The big question is: how do we proceed with such log changes which could be considered breaking? I don't think it's a healthy practice to keep adding flags for each and every individual log field we add 🤔

maxmoehl:
> sadly this is a breaking change (whomp)

That's what I was expecting 😄 

> Develop a more advanced config property that allows users to "opt in" to certain fields by name 1 by 1. So that way we can add new fields, but they can always opt in / out of certain ones. - Fixes this problem for the future. But this is more work. And this is just a config property for everything! Will operators even opt in to any of them!?

That's probably what I will be going for in this PR to prevent accumulating further config flags. Just a flat list, and if anyone needs special properties they can be added and exposed via this mechanism without breaking anyone.

> add an additional access-log file that's json-based for people to migrate to, and we deprecate the old one eventually. When it is json based we can add new properties without it being a breaking change. - Personally I love this idea, but it is definitely more work.

This is my preferred long-term solution, but it will require quite some effort so it's not clear when we will be able to invest.

accesslog/schema/access_log_record_test.go
```diff
@@ -314,6 +314,27 @@ var _ = Describe("AccessLogRecord", func() {
 				Eventually(r).Should(Say(`x_cf_routererror:"-"`))
 			})
 		})
+
+		Context("when extra_fields is set to [local_address]", func() {
```
ameowlia, line 318: Can you add an test for...
* when extra_fields does not contain local_address, then the local address is not tested.
* when extra_fields contains "meow" it ... does not error? logs? idk does something reasonable?

ameowlia (changes requested):
❓ Can you add a PR to routing-release so I can deploy and test everything at once?

maxmoehl:
I've opened cloudfoundry/routing-release#448 as a draft as I'm still adding tests.

accesslog/schema/access_log_record_test.go
maxmoehl: I've added two more test cases.

maxmoehl:
The past weeks have been a bit busy, I finally found the time to clean this up. The [rr PR](https://github.com/cloudfoundry/routing-release/pull/448) is also ready now. Unfortunately test execution on my end is still hindered by cloudfoundry/routing-release#450. I ran the package tests using `go test` and they seem fine but I can't run the whole shebang, could you please do that? Thank you!

I'm currently also deploying this to one of our environments.

accesslog/schema/access_log_record.go
```diff
@@ -319,6 +322,17 @@ func (r *AccessLogRecord) makeRecord(performTruncate bool) []byte {
 		b.WriteDashOrFloatValue(r.successfulAttemptTime())
 	}
 
+	// We have to consider the impact of iterating over a list. This technically allows to repeat
+	// some of the fields but it allows us to iterate over the list only once instead of once per
+	// field when we perform a [slices.Contains] check.
+	for _, field := range r.ExtraFields {
```
ameowlia, line 328: If we iterate over the list this means the operator will get to determine the order of these, vs a standardized order. 

Does that matter?

proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -191,6 +191,7 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
 				slog.Float64("dns-lookup-time", trace.DnsTime()),
 				slog.Float64("dial-time", trace.DialTime()),
 				slog.Float64("tls-handshake-time", trace.TlsTime()),
+				slog.String("local-addr", trace.LocalAddr()),
```
ameowlia, line 194: ❓ Why call it "local-addr" here, but "local_address" above? Why not "local-address" here?

proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -255,6 +256,7 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
...
```
ameowlia, line 259: From reading your spec description on the other PR, it is not clear if the "extra_headers" property is only for access logs or not. 

❓ Based on my reading of it, I would've thought it would affect these logs too? Thoughts?

ameowlia:
✅ tests pass

proxy/round_tripper/proxy_round_tripper.go
maxmoehl: Hm, in the past we didn't give much thought to the compatibility of non-access-logs, primarily because these are JSON and we expect people to parse it as such. Additional fields shouldn't cause any breakage in those cases. For example the timing info directly above this line is controlled by a feature flag in the access logs but unconditionally included in this log.

accesslog/schema/access_log_record.go
maxmoehl: It probably does matter. The other thought I had is that you can end up with multiple instances of the same field (maybe you use some complex templating for the manifest and accidentally add it twice that way). What about this: during the creation of the access logger we de-duplicate the list and order it alphabetically. That way we have a consistent ordering and prevent people from having duplicate fields.

The only concern would then be if someone intentionally wants to influence the order for whatever reason but we usually don't allow that anyway.

proxy/round_tripper/proxy_round_tripper.go
maxmoehl: Fixed it.

accesslog/schema/access_log_record.go
ameowlia: > What about this: during the creation of the access logger we de-duplicate the list and order it alphabetically. That way we have a consistent ordering and prevent people from having duplicate fields.

I like the idea of de-duplicating.

I theoretically like the idea of alphabetizing, but I don't like that I think it would cause breaking changes. I know we only have one "extra field" here right now. But I think eventually we want to put all (most??) access logger fields to be controlled by this property, correct? And if it alphabetizes it, then the order would change.

tl;dr:
✅ deduping
⛔ alphabetizing imo

proxy/round_tripper/proxy_round_tripper.go
ameowlia: > I could adjust the wording in the release spec to reflect that clearly.

Sounds good to me.

accesslog/schema/access_log_record.go
maxmoehl: Fields are now de-duplicated!

ameowlia (approved):

PR #456, opened 2024-12-12
Improve gorouter Route Registry Message Metrics

hoffmaen:

Summary
Solution for https://github.com/cloudfoundry/routing-release/issues/445.

Backward Compatibility
---------------
Breaking Change? **Yes** - metrics and log messages for route registry are renamed (needs to be fixed)


metrics/metricsreporter.go
```diff
@@ -135,27 +135,24 @@ func (m *MetricsReporter) CaptureRoutesPruned(routesPruned uint64) {
...
 	}
 	m.Batcher.BatchIncrementCounter(componentName)
 }
 
-func (m *MetricsReporter) CaptureUnregistryMessage(msg ComponentTagged) {
+func (m *MetricsReporter) CaptureUnregistryMessage(msg ComponentTagged, action string) {
 	var componentName string
 	if msg.Component() == "" {
-		componentName = "unregistry_message"
+		componentName = "unregistry_message." + action
```
peanball, line 151: suggestion to reverse the logic a bit. This suggestion can't be applied all at once because I can't reference the necessary lines in this PR.
```suggestion
componentName := "unregisty_message." + action
if msg.Component() != "" {
    componentName = componentName + "." msg.Component()
}
```

metrics/metricsreporter.go
```diff
@@ -135,27 +135,24 @@ func (m *MetricsReporter) CaptureRoutesPruned(routesPruned uint64) {
...
 	} else {
-		componentName = "unregistry_message." + msg.Component()
-	}
-	err := m.Sender.IncrementCounter(componentName)
-	if err != nil {
-		m.Logger.Debug("failed-sending-metric", log.ErrAttr(err), slog.String("metric", componentName))
+		componentName = "unregistry_message." + action + "." + msg.Component()
 	}
+	m.Batcher.BatchIncrementCounter(componentName)
```
peanball, line 155: is this equivalent to the previously used `m.Sender.IncrementCounter(componentName)`?

registry/registry.go
```diff
@@ -87,27 +87,15 @@ func (r *RouteRegistry) Register(uri route.Uri, endpoint *route.Endpoint) {
...
-		if r.logger.Enabled(context.Background(), slog.LevelDebug) {
-			r.logger.Debug("endpoint-registered", buildSlogAttrs(uri, endpoint)...)
-		}
-	default:
-		if r.logger.Enabled(context.Background(), slog.LevelDebug) {
-			r.logger.Debug("endpoint-not-registered", buildSlogAttrs(uri, endpoint)...)
-		}
+	r.reporter.CaptureRegistryMessage(endpoint, string(endpointPutResult))
+	if r.logger.Enabled(context.Background(), slog.LevelInfo) {
+		r.logger.Info(string(endpointPutResult), buildSlogAttrs(uri, endpoint)...)
```
peanball, line 98: this will significantly change the log output, as we'll have log messages for each updated (including no-op) and 'not-registered' messages, right?

registry/registry.go
```diff
@@ -168,42 +156,54 @@ func (r *RouteRegistry) Unregister(uri route.Uri, endpoint *route.Endpoint) {
...
-				r.logger.Info("route-unregistered", slog.Any("uri", uri))
+	if pool.IsEmpty() {
+		if r.EmptyPoolResponseCode503 && r.EmptyPoolTimeout > 0 {
+			if time.Since(pool.LastUpdated()) > r.EmptyPoolTimeout {
+				r.byURI.Delete(routeKey)
+				return route.RouteUnregistered
 			}
+		} else {
+			r.byURI.Delete(routeKey)
+			return route.RouteUnregistered
```
peanball, line 203:
```suggestion
		if !(r.EmptyPoolResponseCode503 && r.EmptyPoolTimeout > 0) 
		    || (r.EmptyPoolResponseCode503 && r.EmptyPoolTimeout > 0 
		        && if time.Since(pool.LastUpdated()) > r.EmptyPoolTimeout) {
				r.byURI.Delete(routeKey)
				return route.RouteUnregistered
			}
```

peanball (dismissed):
most important comment: route registration logging for any result is now `Info` level and will cause significant increase in logs in production. Please change it back to debug level output for "refreshes" and "no-ops".

Some other smaller suggestions are there but not vital.

metrics/metricsreporter.go
```diff
@@ -135,29 +135,16 @@ func (m *MetricsReporter) CaptureRoutesPruned(routesPruned uint64) {
...
-		componentName = "unregistry_message"
-	} else {
-		componentName = "unregistry_message." + msg.Component()
-	}
-	err := m.Sender.IncrementCounter(componentName)
-	if err != nil {
-		m.Logger.Debug("failed-sending-metric", log.ErrAttr(err), slog.String("metric", componentName))
-	}
-}
-
```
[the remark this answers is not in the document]
peanball: just for my understanding:

Would you then want to have the "old" total metric and new breakdown metrics in addition?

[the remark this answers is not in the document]
maxmoehl: What I mean is that this change:

https://github.com/cloudfoundry/gorouter/pull/456/files#diff-5593a38c8a92809ab9448ee33f120e244e7fc4c0fd0689e239a9e85dc4ae0933L173-R188

will change the emitted metrics for unregister messages from `unregistry_message.*` to `registry_message.*`. This would be a breaking change which doesn't seem to provide much value and could be avoided.

metrics/metricsreporter.go
b1tamara: Applied

registry/registry.go
b1tamara: We changed it. Only changes will be logged with info level.

registry/registry.go
b1tamara: Applied

metrics/metricsreporter.go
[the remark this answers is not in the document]
b1tamara: You are right. We changed it back.

metrics/metricsreporter.go
```diff
@@ -135,27 +135,22 @@ func (m *MetricsReporter) CaptureRoutesPruned(routesPruned uint64) {
...
 
-func (m *MetricsReporter) CaptureRegistryMessage(msg ComponentTagged) {
+func (m *MetricsReporter) CaptureRegistryMessage(msg ComponentTagged, action string) {
 	var componentName string
 	if msg.Component() == "" {
-		componentName = "registry_message"
+		componentName = "registry_message." + action
 	} else {
-		componentName = "registry_message." + msg.Component()
+		componentName = "registry_message." + action + "." + msg.Component()
```
maxmoehl, line 143: This will probably also change the metric right? I was hoping that we could preserve the metric name but only add an additional tag to it.

maxmoehl: [This](https://github.com/cloudfoundry/dropsonde/blob/cce4f49c37d11195c32193cc23fe09210d4ca18c/metricbatcher/metricbatcher.go#L195) is what I had in mind. But it seems like the API is a bit cumbersome to use...

PR #457, opened 2024-12-12
SurgicalPool extension by the load balancing algorithm (the same way as the RoutePool)

Dariquest:
## Summary

Issue https://github.com/cloudfoundry/gorouter/issues/458

LoadBalancing Algorithm was introduced for the RoutePool in the [PR](https://github.com/cloudfoundry/gorouter/pull/428). 
The load balancing algorithm is also needed to be passed to the SurgicalPool (included in the RouteOpts for the initialization).
Missing load balancing algorithm in the surgical pool leads to the error "invalid-pool-load-balancing-algorithm". 
This error is intended validation for a pool load balancing algorithm. An additional log parameter was added for this error either.

## Reproduction Steps

When I curl an app with an x-cf-app-instance header

curl -v "https://example-app.com/" -H "x-cf-app-instance: app_id:instance_id"

The same curl without the instance header works properly.

## Backward Compatibility
---------------
Breaking Change? **No**


Dariquest:
Add label to an issue https://github.com/cloudfoundry/gorouter/issues/458

route/pool.go
```diff
@@ -384,7 +384,9 @@ func (p *EndpointPool) Endpoints(logger *slog.Logger, initial string, mustBeStic
 		logger.Debug("endpoint-iterator-with-round-robin-lb-algo")
 		return NewRoundRobin(logger, p, initial, mustBeSticky, azPreference == config.AZ_PREF_LOCAL, az)
 	default:
-		logger.Error("invalid-pool-load-balancing-algorithm", slog.String("poolLBAlgorithm", p.LoadBalancingAlgorithm))
+		logger.Error("invalid-pool-load-balancing-algorithm",
+			slog.String("poolLBAlgorithm", p.LoadBalancingAlgorithm),
+			slog.String("poolRouteSvcUrl", p.RouteSvcUrl))
```
peanball, line 389: we should have the pool's URL here, not the route service URL. It's not a property of `EndpointPool` unfortunately, but only recorded in a map of `route_url -> EndpointPool`.

Dariquest: done

route/pool.go
```diff
@@ -384,7 +384,9 @@ func (p *EndpointPool) Endpoints(logger *slog.Logger, initial string, mustBeStic
 		logger.Debug("endpoint-iterator-with-round-robin-lb-algo")
 		return NewRoundRobin(logger, p, initial, mustBeSticky, azPreference == config.AZ_PREF_LOCAL, az)
 	default:
-		logger.Error("invalid-pool-load-balancing-algorithm", slog.String("poolLBAlgorithm", p.LoadBalancingAlgorithm))
+		logger.Error("invalid-pool-load-balancing-algorithm",
+			slog.String("poolLBAlgorithm", p.LoadBalancingAlgorithm),
+			slog.Group("data", slog.String("Host", p.host)))
```
peanball, line 389: Sorry, I made a bad suggestion yesterday.

The additional `slog.*` data in the `logger.Error` is already wrapped in a `data` group by default. Correcting the suggestion. But please test and make sure that the log looks reasonable, i.e. something like this:

```
{"log_level": 1, "message": "invalid-pool-load-balancing-algorithm", ...,
 "data": { "Host":"the.pool.host", "poolLBAlgorithm", "round-robin" } }
```

You may need to undo your fix temporarily in the test environment and force the condition again with the instance-pinned request.

```suggestion
			slog.String("Host", p.host))
```

Dariquest:
Replacing this PR by https://github.com/cloudfoundry/gorouter/pull/460

Issue #458, opened 2024-12-13
SurgicalPool has no load balancing algorithm leading to "invalid-pool-load-balancing-algorithm" error


Dariquest:
### Current behavior

LoadBalancing Algorithm was introduced for the RoutePool in the https://github.com/cloudfoundry/gorouter/pull/428.
The load balancing algorithm was not passed to the SurgicalPool (included in the RouteOpts for the initialization).

Missing load balancing algorithm in the surgical pool leads to the error "invalid-pool-load-balancing-algorithm".
This error is intended validation for a pool load balancing algorithm. 
An additional log parameter needs to be added for this error for error research simplification.

### Desired behavior

The load balancing algorithm is also needed to be passed to the SurgicalPool (included in the RouteOpts for the initialization).

### Reproduction Steps

When I curl an app with an x-cf-app-instance header

`curl -v "https://example-app.com/" -H "x-cf-app-instance: app_id:instance_id"`

The same curl without the instance header works properly.

### Impact
 
The custom per-route load balancing algorithm feature is not finally released. 
This error has no functional impact in routing behaviour.  

### Affected Version

#428
The error occurs since routing-release [version](https://github.com/cloudfoundry/routing-release/releases/tag/v0.302.0)

winkingturtle-vmw:
Released in 0.327.0

PR #460, opened 2024-12-17
SurgicalPool extension by the load balancing algorithm (the same way as the RoutePool)

Dariquest:
## Summary

Issue https://github.com/cloudfoundry/gorouter/issues/458

LoadBalancing Algorithm was introduced for the RoutePool in the [PR](https://github.com/cloudfoundry/gorouter/pull/428). 
The load balancing algorithm is also needed to be passed to the SurgicalPool (included in the RouteOpts for the initialization).
Missing load balancing algorithm in the surgical pool leads to the error "invalid-pool-load-balancing-algorithm". 
This error is intended validation for a pool load balancing algorithm. An additional log parameter was added for this error either.

## Reproduction Steps

When I curl an app with an x-cf-app-instance header

curl -v "https://example-app.com/" -H "x-cf-app-instance: app_id:instance_id"

The same curl without the instance header works properly.

## Backward Compatibility
---------------
Breaking Change? **No**

### DIFF
--- a/registry/registry.go
+++ b/registry/registry.go
@@ -267,6 +267,7 @@ func (r *RouteRegistry) LookupWithInstance(uri route.Uri, appID string, appIndex
 				Host:               p.Host(),
 				ContextPath:        p.ContextPath(),
 				MaxConnsPerBackend: p.MaxConnsPerBackend(),
+				LoadBalancingAlgorithm: p.LoadBalancingAlgorithm,
 			})
 			surgicalPool.Put(e)
 		}


Dariquest:
Adding reference to an issue https://github.com/cloudfoundry/gorouter/issues/458

ameowlia:
Please make sure you run ALL of the tests via `routing-release/scripts/test-in-docker`. 

This PR broke the tests because of a go fmt issue that would've been caught with the script. Also I am concerned that there were no tests for this PR...

Dariquest:
Hi Amelia @ameowlia,
thanks for you comment and sorry for breaking the format and missing tests. 
The unit tests for the surgical pool case are provided in a new PR https://github.com/cloudfoundry/gorouter/pull/462

PR #435, opened 2024-09-03
Replace zap with slog; Replace logging wrappers with slog 

hoffmaen:
- [x] Read the [Contributing document](../blob/-/.github/CONTRIBUTING.md).

Summary
---------------

gorouter uses a very outdated `zap` version from 2016 for logging. This Pull-Request introduces required code changes to use `slog` as the logging frontend, together with a zap handler via `zapslog`. Slog provides a stable frontend, while we keep using zap with its configuration as the backend. 

Both the gorouter logger and the lager logger (used for initializing the UAA Token Fetcher) used zap logger wrappers. This Pull-Request replaces the logger wrapper with direct use of `slog.Logger`. A thin library is provided to keep it compatible with the code. The library provides the functions `CreateLogger` and `CreateLoggerWithSource` to create a logger that subsequently writes data into the `data` field, and optionally sets the `source` field in the log's root. The library also has functions to dynamically set the `WriteSyncer`, which is required for tests, and to dynamically set the logging level and timestamp format.

With the removal of the `Logger` interface, all occurrences of the FakeLogger have been replaced with an `slog.Logger` based `TestLogger`.

Additionally, some test refactoring and code reformatting has been done.

Backward Compatibility
---------------
Breaking Change? **No**

This Pull-Request requires https://github.com/cloudfoundry/routing-release/pull/434 for updating zap dependencies.

The zap handler configuration has been kept, so the log message format does not change. Workarounds for differing behavior of the `slog.Logger` frontend have been found and implemented. There are quite a lot of tests on log messages in the code, an additional end-to-end test with a deployment of this code did not show any differences compared to the old implementation.

PR #464, opened 2025-01-31
fix(slog): disable automatic logging of stacktraces for errors

peanball:
The slog handler for Zap is configured to automatically emit stack traces at "Error" level.

In pure Zap, stack traces are added via option AddStacktrace, which was not used before the move to slog.

The Zap slog handler has a mapping from Zap levels to slog levels, and does not emit values above slog.LevelError. By setting the limit for automatically added stack traces via slog to ErrorLevel + 1, this effectively disables automatic stack traces.

The `panic-check` handler that catches and logs panics in handlers remains unaffected, as it explicitly logs a field called "stacktrace", as it did with pure Zap.

Related: #435

Breaking change: No. This is restoring the behavior we had with the pure zap logger / logger wrappers.

### DIFF
--- a/handlers/paniccheck.go
+++ b/handlers/paniccheck.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
-	"runtime"
+	"runtime/debug"
 
 	"github.com/urfave/negroni/v3"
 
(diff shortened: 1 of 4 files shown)


maxmoehl (changes requested):
I was curious because you mentioned the panic-check handler, seems like there's a bug in there:

https://github.com/cloudfoundry/gorouter/blob/a34e69090fca2e908d86a2982b98d0f46e07aa6c/handlers/paniccheck.go#L44

From [`runtime.StartTrace`](https://pkg.go.dev/runtime#StartTrace):

> StartTrace enables tracing for the current process.

that doesn't sound right. It should probably call [`runtime/debug.Stack`](https://pkg.go.dev/runtime/debug#Stack). Without your change this is not noticed because the stack-trace is emitted either way, but by merging this there will be no stack-trace anymore. And if you don't mind, a test to ensure the panic-check handler emits the stack trace would probably be a good idea as well 😅

peanball:
+100. Will do.

peanball:
@maxmoehl, I've applied the change you mention and ensure in the `paniccheck_test.go` that the stack trace is actually there. I'm checking that an expected fragment that should show up in the stacktrace is really there.

maxmoehl (approved):

PR #466, opened 2025-02-17
Prometheus metrics

mike-jc:

Summary
---------------
- We want to migrate from EnvelopeV1 metrics to Prometheus (and OpenTelemetry) ones
          - The existing Prometheus metric (HTTP Latency) were moved to the new prom metrics service. Its histogram
             buckets are exposed to routing release as config parameter (current values as default). 
- We extended go-metric-register https://github.com/cloudfoundry/go-metric-registry/pull/146

Backward Compatibility
---------------
Breaking Change? **No**


main.go
```diff
@@ -257,8 +286,8 @@ func main() {
 	monitor := ifrit.Invoke(sigmon.New(group, syscall.SIGTERM, syscall.SIGINT, syscall.SIGUSR1))
 
 	go func() {
-		time.Sleep(c.RouteLatencyMetricMuzzleDuration) // this way we avoid reporting metrics for pre-existing routes
-		metricsReporter.UnmuzzleRouteRegistrationLatency()
+		time.Sleep(c.RouteLatencyMetricMuzzleDuration)     // this way we avoid reporting metrics for pre-existing routes
+		registryMetrics.UnmuzzleRouteRegistrationLatency() // TODO: look at this
```
b1tamara, line 290: What is about this TODO?

metrics_prometheus/metrics.go
```diff
@@ -0,0 +1,198 @@
...
+
+func (metrics *Metrics) CaptureEmptyContentLengthHeader() {
+	metrics.EmptyContentLengthHeader.Add(1)
+}
+
+// CaptureRoutingRequest used to capture backend round trips
+func (metrics *Metrics) CaptureRoutingRequest(b *route.Endpoint) {
+	component := b.Component()
+	labels := []string{component, "no"} // component, is-routed-app
+	if strings.HasPrefix(component, "dea-") {
```
b1tamara, line 159: Do we want to keep the specific logic related to "dea"-component for prometheus?

metrics_prometheus/metrics_test.go
```diff
@@ -0,0 +1,435 @@
...
+			var config = config.PrometheusConfig{Port: 0, Meters: getMetersConfig()}
+			r = NewMetricsRegistry(config)
+			m = NewMetrics(r, perRequestMetricsReporting, config.Meters)
+			endpoint = new(route.Endpoint)
+		})
+		AfterEach(func() {
+			m.perRequestMetricsReporting = true
+		})
+
+		It("sends number of nats messages received from each component", func() {
```
b1tamara, line 35: The tests for CaptureUnregisterMessage are missing.

mike-jc: Added

main.go
mike-jc: We still need that call while we support Envelope V1. Added explanation in comment

metrics_prometheus/metrics.go
mike-jc: Good call ! We do not want to keep DEA support in Prometheus metrics. Removed

metrics/metricsreporter.go
```diff
@@ -15,6 +15,11 @@ import (
 	"github.com/cloudfoundry/dropsonde/metrics"
 )
 
+var _ interface {
+	ProxyReporter
+	RouteRegistryReporter
+} = &MetricsReporter{}
+
```
maxmoehl, line 22: These are not required, I'll leave it up to you whether you would prefer to keep them or remove them.

metrics_prometheus/metrics.go
```diff
@@ -0,0 +1,193 @@
...
+	RouteRegistration           mr.CounterVec
+	RouteUnregistration         mr.CounterVec
+	RoutesPruned                mr.Counter
+	TotalRoutes                 mr.Gauge
+	TimeSinceLastRegistryUpdate mr.Gauge
+	RouteLookupTime             mr.Histogram
+	RouteRegistrationLatency    mr.Histogram
+	RoutingRequest              mr.CounterVec
+	BadRequest                  mr.Counter
+	BadGateway                  mr.Counter // TODO from b1tamara: rename to BackendBadGateway?
```
maxmoehl, line 27: TODO

metrics_prometheus/metrics.go
```diff
@@ -0,0 +1,193 @@
...
+
+func (metrics *Metrics) CaptureRouteRegistrationLatency(t time.Duration) {
+	metrics.RouteRegistrationLatency.Observe(float64(t) / float64(time.Millisecond))
+}
+
+// UnmuzzleRouteRegistrationLatency should set a flag which suppresses metric data.
+// That makes sense for Envelope V1 where we send it to collector any time we got new value
+// but is unnecessary for Prometheus where data is buffered and sent to collector on constant frequency base.
+// We still need this method though to fulfil the interface.
+func (metrics *Metrics) UnmuzzleRouteRegistrationLatency() {}
```
maxmoehl, line 125: I'm still not sure if we should keep the logic from the envelopes here. Maybe @ameowlia knows this or whom we could ask about this feature?

registry/registry.go
```diff
@@ -45,7 +45,7 @@ type RouteRegistry struct {
 	pruneStaleDropletsInterval time.Duration
 	dropletStaleThreshold      time.Duration
 
-	reporter metrics.RouteRegistryReporter
+	multiReporter metrics.MultiRouteRegistryReporter
```
maxmoehl, line 48: Couldn't this type stay the interface type it was previously? The coding here should not need to know what reporter implementation is passed to it.

registry/registry.go
```diff
@@ -109,6 +109,7 @@ func (r *RouteRegistry) Register(uri route.Uri, endpoint *route.Endpoint) {
 			r.logger.Debug("endpoint-not-registered", buildSlogAttrs(uri, endpoint)...)
 		}
 	}
+
```
maxmoehl, line 112: Empty line.

registry/registry.go
```diff
@@ -87,15 +87,15 @@ func (r *RouteRegistry) Register(uri route.Uri, endpoint *route.Endpoint) {
 		return
 	}
 
-	endpointAdded := r.register(uri, endpoint)
+	result := r.register(uri, endpoint)
```
maxmoehl, line 90: (I know this change was made by me) We should probably put this back and defer that to the round of clean-up we plan to do afterwards.

maxmoehl (commented):
When we merge the PR, we should make to squash all the commits.

metrics_prometheus/metrics.go
mike-jc: Do you think of removing this totally (from envelopes as well) ? Otherwise we still need this for ProxyReporter interface.

metrics/metricsreporter.go
mike-jc: Removed it

metrics_prometheus/metrics.go
maxmoehl: I meant replicating the logic we have for the envelopes for prometheus as well instead of making it a no-op.

mike-jc: Per @ameowlia answer in work group: we keep this as it is, supporting unmuzzling for envelopes and keep support for envelopes and prometheus for now

maxmoehl (approved):

geofffranks:
I took a cursory look at this, and it looks good. I don't see any changes that would end up breaking things without operators opting in, or causing issues with edge case configurations.

This will still need a full review from an ARP approver, running tests, ensuring deployment works as intended, and validation of doc updates for the new prometheus metrics.

metrics/compositereporter.go
```diff
@@ -44,17 +44,146 @@ type RouteRegistryReporter interface {
 	CaptureRouteStats(totalRoutes int, msSinceLastUpdate int64)
 	CaptureRoutesPruned(prunedRoutes uint64)
 	CaptureLookupTime(t time.Duration)
-	CaptureRegistryMessage(msg ComponentTagged)
+	CaptureRegistryMessage(msg ComponentTagged, action string)
 	CaptureRouteRegistrationLatency(t time.Duration)
-	UnmuzzleRouteRegistrationLatency()
 	CaptureUnregistryMessage(msg ComponentTagged)
+	// Deprecated: used only in Envelope v1, should be removed when Envelope v1 support is eliminated
```
peanball, line 50:
```suggestion
	// Deprecated: used only in Envelope v1, should be removed, if Envelope v1 support is removed
```

main.go
```diff
@@ -130,17 +132,51 @@ func main() {
 
 	}
 
+	// setup metrics via dropsonse
 	sender := metric_sender.NewMetricSender(dropsonde.AutowiredEmitter())
+	var metricsReporter *metrics.MetricsReporter
+	if c.EnableEnvelopeV1Metrics {
```
peanball, line 138: This still creates the dropsonde emitter regardless of that setting. 

The dropsonde `sender` is also still used in the FD monitor and the NATS monitor. Is that intended?

main.go
```diff
@@ -130,17 +132,51 @@ func main() {
...
+
+	// setup metrics via prometheus
+	var registryMetrics metrics.MultiRouteRegistryReporter
+
+	if metricsReporter != nil {
+		registryMetrics = append(registryMetrics, metricsReporter)
+	}
+	if promMetrics != nil {
+		registryMetrics = append(registryMetrics, promMetrics)
+	}
```
peanball, line 157: You could do the initialization of dropsonde, or prometheus, or both in a separate function that then returns a registry reporter, either a "regular" one or the `MultiRouteRegistryReporter`.

main.go
```diff
@@ -130,17 +132,51 @@ func main() {
...
+
+	var proxyMetrics metrics.MultiProxyReporter
+
+	if metricsReporter != nil {
+		proxyMetrics = append(proxyMetrics, metricsReporter)
+	}
+
+	if promMetrics != nil {
+		proxyMetrics = append(proxyMetrics, promMetrics)
+	}
```
peanball, line 171: Same here. Alternatively you pull the whole thing out and just return the `CompositeReporter`.

main.go
```diff
@@ -130,17 +132,51 @@ func main() {
...
-	metricsReporter := initializeMetrics(sender, c, grlog.CreateLoggerWithSource(prefix, "metricsreporter"))
-	fdMonitor := initializeFDMonitor(sender, grlog.CreateLoggerWithSource(prefix, "FileDescriptor"))
-	registry := rregistry.NewRouteRegistry(grlog.CreateLoggerWithSource(prefix, "registry"), c, metricsReporter)
 	if c.SuspendPruningIfNatsUnavailable {
 		registry.SuspendPruning(func() bool { return !(natsClient.Status() == nats.CONNECTED) })
 	}
 
-	varz := rvarz.NewVarz(registry)
-	compositeReporter := &metrics.CompositeReporter{VarzReporter: varz, ProxyReporter: metricsReporter}
+	fdMonitor := initializeFDMonitor(sender, grlog.CreateLoggerWithSource(prefix, "FileDescriptor"))
```
peanball, line 179: this goes purely to the dropsonde metrics, not prometheus.

Same for the NATS monitor.

metrics/compositereporter.go
```diff
@@ -44,17 +44,146 @@ type RouteRegistryReporter interface {
...
 }
 
 type CompositeReporter struct {
 	VarzReporter
 	ProxyReporter
 }
 
+type MultiRouteRegistryReporter []RouteRegistryReporter
+
+var _ RouteRegistryReporter = MultiRouteRegistryReporter{}
```
peanball, line 61: Can you explain the blank identifier and assignment?

Is that to register the interface implementation?

metrics_prometheus/metrics.go
```diff
@@ -0,0 +1,193 @@
...
+		metrics.RouteLookupTime.Observe(float64(t.Nanoseconds()))
+	}
+}
+
+func (metrics *Metrics) CaptureRouteRegistrationLatency(t time.Duration) {
+	metrics.RouteRegistrationLatency.Observe(float64(t) / float64(time.Millisecond))
+}
+
+// UnmuzzleRouteRegistrationLatency should set a flag which suppresses metric data.
+// Deprecated: that makes sense for Envelope V1 where we send it to collector any time we got new value
```
peanball, line 122: Please review the wording. As far as I understood, we don't deprecated Envelope V1 metrics overall for Gorouter. While some setups might not want to use them, they remain supported.

metrics_prometheus/metrics.go
```diff
@@ -0,0 +1,193 @@
...
+	WebsocketFailures           mr.Counter
+	Responses                   mr.CounterVec
+	RouteServicesResponses      mr.CounterVec
+	RoutingResponseLatency      mr.HistogramVec
+	perRequestMetricsReporting  bool
+}
+
+func NewMetricsRegistry(config config.PrometheusConfig) *mr.Registry {
+	var metricsRegistry *mr.Registry
+	if config.Port != 0 && config.CertPath != "" {
```
peanball, line 43: This allows `config.Port == 0` when cert path is not set, which would start on a random port?

Either way the distinction for TLS or not should be on CertPath, not on port, I think.

peanball (commented):
Some questions and some wording things.

I'm not sure if the NATS and FD metrics would be in scope / useful in Prometheus too. Currently they seem to be only in Envelope v1 / Dropsonde.

This also means you can't disable dropsonde entirely with this PR.

metrics/compositereporter.go
maxmoehl: See https://github.com/cloudfoundry/gorouter/pull/466#discussion_r1962944351, it's an easy way to check whether you implemented the interface correctly as some IDE inspections will offer to complete the implementation.

metrics_prometheus/metrics.go
b1tamara: We are using port = 0 in our tests , in order to start the test on the random port. Otherwise it is not possible to run the tests.
In the routing release we added a check which does not allow to set port = 0.

main.go
```diff
@@ -294,6 +264,28 @@ func main() {
 	os.Exit(0)
 }
 
+// initializeDropsondeReporter setups metrics via dropsonse if enabled
+func initializeDropsondeReporter(prefix string, logger *slog.Logger, c *config.Config) *metrics.MetricsReporter {
+	if !c.EnableEnvelopeV1Metrics {
+		return nil
+	}
+	err := dropsonde.Initialize(c.Logging.MetronAddress, c.Logging.JobName)
+	if err != nil {
```
b1tamara, line 273: Do we need the return value in this case?

main.go
```diff
@@ -294,6 +264,28 @@ func main() {
...
+		grlog.Fatal(logger, "dropsonde-initialize-error", grlog.ErrAttr(err))
+	}
+	dropsondeMetricSender := metric_sender.NewMetricSender(dropsonde.AutowiredEmitter())
+	return initializeMetrics(dropsondeMetricSender, c, grlog.CreateLoggerWithSource(prefix, "metricsreporter"))
```
b1tamara, line 277: I find it more readable to store the value in a variable and then return the variable.

main.go
mike-jc: Fatal() on the next line makes os.Exit(1) and stops the entire process

main.go
mike-jc: I know that it's more or less standard to not introduce a new variable if we are not going to use it. But am ok with the opposite approach as well, for the sake of readability. Just would like to know if it is a community approach?

peanball:
With this PR, and https://github.com/cloudfoundry/routing-release/pull/456 I have a crash when not configuring anything for Prometheus:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x955153]

goroutine 28 [running]:
code.cloudfoundry.org/gorouter/metrics_prometheus.(*Metrics).CaptureRegistryMessage(0x0, {0xe31b20?, 0xc00026c780?}, {0xcfe106, 0x5})
        /var/vcap/data/compile/gorouter/src/code.cloudfoundry.org/gorouter/metrics_prometheus/metrics.go:100 +0x93
code.cloudfoundry.org/gorouter/metrics.MultiRouteRegistryReporter.CaptureRegistryMessage(...)
        /var/vcap/data/compile/gorouter/src/code.cloudfoundry.org/gorouter/metrics/compositereporter.go:112
code.cloudfoundry.org/gorouter/registry.(*RouteRegistry).Register(0xc00017f5e0, {0xc0001245c0, 0x36}, 0xc00026c780)
        /var/vcap/data/compile/gorouter/src/code.cloudfoundry.org/gorouter/registry/registry.go:92 +0x1bf
code.cloudfoundry.org/gorouter/mbus.(*Subscriber).registerEndpoint(0xc000200b60, 0xc0005ce460)
        /var/vcap/data/compile/gorouter/src/code.cloudfoundry.org/gorouter/mbus/subscriber.go:252 +0x218
code.cloudfoundry.org/gorouter/mbus.(*Subscriber).subscribeRoutes.func1(0xc000214540)
        /var/vcap/data/compile/gorouter/src/code.cloudfoundry.org/gorouter/mbus/subscriber.go:221 +0xac
github.com/nats-io/nats%2ego.(*Conn).waitForMsgs(0xc00025a708, 0xc0005d41e0)
        /var/vcap/data/compile/gorouter/src/code.cloudfoundry.org/vendor/github.com/nats-io/nats.go/nats.go:3195 +0x47d
created by github.com/nats-io/nats%2ego.(*Conn).subscribeLocked in goroutine 24
        /var/vcap/data/compile/gorouter/src/code.cloudfoundry.org/vendor/github.com/nats-io/nats.go/nats.go:4441 +0x41b
```

main.go
```diff
@@ -130,17 +125,21 @@ func main() {
 
 	}
 
-	sender := metric_sender.NewMetricSender(dropsonde.AutowiredEmitter())
+	dropReporter := initializeDropsondeReporter(prefix, logger, c)
+	promReporter := initializePrometheusReporter(c)
+	registryMetrics := metrics.NewMultiRouteRegistryReporter(dropReporter, promReporter)
+	proxyMetrics := metrics.NewMultiProxyReporter(dropReporter, promReporter)
+	monitorMetrics := metrics.NewMultiMonitorReporter(dropReporter, promReporter)
```
peanball, line 132: Please make sure that the list of reporters for the various `*Multi*Reporter`s just doesn't contain nil values by passing a list, not 2 args. Should make it simpler and less error prone.

metrics_prometheus/metrics.go
```diff
@@ -0,0 +1,214 @@
...
+	metrics.RoutesPruned.Add(float64(routesPruned))
+}
+
+func (metrics *Metrics) CaptureTotalRoutes(totalRoutes int) {
+	metrics.TotalRoutes.Set(float64(totalRoutes))
+}
+
+func (metrics *Metrics) CaptureTimeSinceLastRegistryUpdate(msSinceLastUpdate int64) {
+	metrics.TimeSinceLastRegistryUpdate.Set(float64(msSinceLastUpdate))
+}
```
peanball, line 114: This is only used in a test.

mike-jc: Removed this. The metric is used in CaptureRouteStats() method

peanball (approved):
Verified that metrics are generated in Prometheus, prom_scraper forwards them, etc. successfully.

Also no change / breaking for the default configuration when prometheus is off.

peanball:
> When we merge the PR, we should make sure to squash all the commits.

Sorry, I completely forgot about that :( It's merged as is now.

PR #470, opened 2025-03-28
Retry new endpoint backends if they are being added while serving the request

mariash:

Summary
---------------
Support application availability during rolling restarts. During application restart with rolling strategy new instance comes up first and then old instance is removed. If the number of application instances is less than the number of attempts gorouter might not try to serve new instance if it comes up while request is being served. This is because it saves in memory number of instances and only retries up to that number. 

Instead retry next instance in the pool and only leave the loop before the maximum number of attempts if there were no updates to the pool.
When new instance is registered pool gets updated.
When instance is removed shift all instances in the pool instead of moving last registered element in its place. This way last registered element will stay at the end of the pool and loop will get to it faster.

Backward Compatibility
---------------
Breaking Change? **No**

### DIFF
--- a/proxy/round_tripper/proxy_round_tripper.go
+++ b/proxy/round_tripper/proxy_round_tripper.go
@@ -137,10 +137,11 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
 	var selectEndpointErr error
 	var maxAttempts int
 	if reqInfo.RouteServiceURL == nil {
-		maxAttempts = max(min(rt.config.Backends.MaxAttempts, reqInfo.RoutePool.NumEndpoints()), 1)
+		maxAttempts = max(rt.config.Backends.MaxAttempts, 1)
 	} else {
 		maxAttempts = rt.config.RouteServiceConfig.MaxAttempts
 	}
+	triedEndpoints := map[string]bool{}
 
 	for attempt := 1; attempt <= maxAttempts; attempt++ {
 		logger := rt.logger
@@ -151,12 +152,25 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
 		if reqInfo.RouteServiceURL == nil {
 			// Because this for-loop is 1-indexed, we substract one from the attempt value passed to selectEndpoint,
 			// which expects a 0-indexed value
-			endpoint, selectEndpointErr = rt.selectEndpoint(iter, request, attempt-1)
+			endpoint, selectEndpointErr = rt.selectEndpoint(iter, attempt-1)
+
+			if attempt > 1 {
+				if attempt > reqInfo.RoutePool.NumEndpoints() {
+					// check if new endpoints were registered
+					if selectEndpointErr == nil {
+						if _, found := triedEndpoints[endpoint.CanonicalAddr()]; found {
+							break
+						}
+					}
+				}
+			}
+
 			if selectEndpointErr != nil {
 				logger.Error("select-endpoint-failed", slog.String("host", reqInfo.RoutePool.Host()), log.ErrAttr(selectEndpointErr))
 				break
 			}
 			logger = logger.With(slog.Group("route-endpoint", endpoint.ToLogData()...))
+			triedEndpoints[endpoint.CanonicalAddr()] = true
 			reqInfo.RouteEndpoint = endpoint
 
 			logger.Debug("backend", slog.Int("attempt", attempt))
@@ -399,7 +413,7 @@ func (rt *roundTripper) timedRoundTrip(tr http.RoundTripper, request *http.Reque
 	return resp, err
 }
 
-func (rt *roundTripper) selectEndpoint(iter route.EndpointIterator, request *http.Request, attempt int) (*route.Endpoint, error) {
+func (rt *roundTripper) selectEndpoint(iter route.EndpointIterator, attempt int) (*route.Endpoint, error) {
 	endpoint := iter.Next(attempt)
 	if endpoint == nil {
 		return nil, NoEndpointsAvailable
--- a/route/pool.go
+++ b/route/pool.go
@@ -7,6 +7,7 @@ import (
 	"maps"
 	"math/rand"
 	"net/http"
+	"slices"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -379,11 +380,10 @@ func (p *EndpointPool) Remove(endpoint *Endpoint) bool {
 func (p *EndpointPool) removeEndpoint(e *endpointElem) {
 	i := e.index
 	es := p.endpoints
-	last := len(es)
-	// re-ordering delete
-	es[last-1], es[i], es = nil, es[last-1], es[:last-1]
-	if i < last-1 {
-		es[i].index = i
+
+	es = slices.Delete(es, i, i+1)
+	for j := i; j < len(es); j++ {
+		es[j].index = j
 	}
 	p.endpoints = es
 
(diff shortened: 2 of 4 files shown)


route/pool.go
```diff
@@ -379,11 +379,10 @@ func (p *EndpointPool) Remove(endpoint *Endpoint) bool {
 func (p *EndpointPool) removeEndpoint(e *endpointElem) {
 	i := e.index
 	es := p.endpoints
-	last := len(es)
-	// re-ordering delete
-	es[last-1], es[i], es = nil, es[last-1], es[:last-1]
-	if i < last-1 {
-		es[i].index = i
+
+	es = append(es[:i], es[i+1:]...)
```
maxmoehl, line 383: Can we use [slices.Delete](https://pkg.go.dev/slices#Delete)? It does some additional cleaning for garbage collection and makes it more clear what is happening here.

proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -163,9 +164,14 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
 		trace.Reset()
 
 		if reqInfo.RouteServiceURL == nil {
+			if attempt > 1 {
+				if attempt > reqInfo.RoutePool.NumEndpoints() && reqInfo.RoutePool.LastUpdated().Equal(lastUpdated) {
+					break
+				}
+			}
```
maxmoehl, line 171: Why do we need the call to `reqInfo.RoutePool.LastUpdated()` here? The first check should always return the current number of endpoints and since `lastUpdated` is never written within the loop it will always be false once the pool was updated, therefore causing an infinite loop.


I'd suggest:
```suggestion
			if attempt > 1 && attempt > reqInfo.RoutePool.NumEndpoints() {
				break
			}
```

mariash: Yes as Amelia mentioned, just checking for number of instances is not enough. We want to account for a case when instance is registered and old instance is removed. This is when the issue is happening, after old instance is removed we fail to connect to it. And during rolling deploy you will have a replacement instance up, so the total number will be the same.

route/pool.go
mariash: Updated

proxy/round_tripper/proxy_round_tripper.go
maxmoehl: Ah I see, from the title I was assuming this is about an app being scaled while gorouter is making requests.

My main point still stands though, assuming the time of the last updated time on the pool changes, `reqInfo.RoutePool.LastUpdated().Equal(lastUpdated)` will _always_ return `false` as we never update the `lastUpdated` which we compare against within the loop and it will always contain the initial value. So, unless I missed something, whenever that happens it would never stop looping.

maxmoehl: I'm trying to come up with a way to fix this but I'm not sure there is a straightforward way. To be on the safe side I wouldn't want to rely on the timestamp as that does't tell us how many instances changed and it could also be the case that there was an instance removed without one being added in which case we wouldn't want to try that endpoint knowing that its gone (if it's one we haven't tried yet).

So I guess without an additional variable we can't make this work. If we had an additional variable, we could use it to store the identity of the endpoints we already tried. When picking the next endpoint the iterator would check whether there is an endpoint we haven't tried and return that.

I've built a small demo but I'm not sure I like it: https://go.dev/play/p/7YuzIoCsbI5 which then outputs:

```
selected endpoint ep1
doesn't work, try again
selected endpoint ep2
doesn't work, try again
selected endpoint ep3
works, stopping
```

I guess what I want to say is that I rather see this as a feature of the iterator, not the round-tripper.

mariash: `reqInfo.RoutePool.LastUpdated().Equal(lastUpdated)` will return true if the pool was updated during the loop run. Whenever we put something in the pool or remove an item we update this value - https://github.com/cloudfoundry/gorouter/blob/6c6bd520ad939910fe8c0830558f6dd179e7d113/route/pool.go#L298

These are all the test case scenarios that cover this (1. not updated, 2. updated with added new backends, 3. updated with 1 added and 1 removed backend) - https://github.com/cloudfoundry/gorouter/blob/8cd1dafe08232ed38614ad00002f3d5bc3d6633b/proxy/round_tripper/proxy_round_tripper_test.go#L407-L472

maxmoehl: But what you don't test is if all backends fail and cause a retry. In that case it would loop indefinitely, no?

mariash: In that case it will retry up to the MaxAttempts which is 3 by default. This is the test for this case - https://github.com/cloudfoundry/gorouter/blob/8cd1dafe08232ed38614ad00002f3d5bc3d6633b/proxy/round_tripper/proxy_round_tripper_test.go#L388-L400

maxmoehl: Sorry, didn't see that 🙈

My main concern is that this is still somewhat of a regression. We have set that limit to a very high value relying on the gorouter not to try any endpoint twice. If you run into overload situations gorouter would keep hammering the few instances which are already overloaded because of the high value. On the other hand, reducing the limit would cause it not to try most of the available endpoints.

mariash: How is it a regression if it retries less than it did before?

tcdowney: @maxmoehl are you saying that you have overwritten the default in your environment to be something very large? https://github.com/cloudfoundry/routing-release/blob/develop/jobs/gorouter/spec#L221

proxy/round_tripper/proxy_round_tripper.go
```diff
@@ -163,9 +164,14 @@ func (rt *roundTripper) RoundTrip(originalRequest *http.Request) (*http.Response
 		trace.Reset()
 
 		if reqInfo.RouteServiceURL == nil {
+			if attempt > 1 {
+				if attempt > reqInfo.RoutePool.NumEndpoints() && reqInfo.RoutePool.LastUpdated().Equal(lastUpdated) {
```
tcdowney, line 168: @maxmoehl wouldn't this prevent it from retrying already-tried endpoints even if MaxAttempts is set to a very large number?

proxy/round_tripper/proxy_round_tripper.go
mariash: @maxmoehl Sorry I was confused, you are right there is an issue when we keep updating the pool while the retries are happening. And thanks to @tcdowney for good questions that helped me to understand the problem.

I added [some test cases](https://github.com/cloudfoundry/gorouter/blob/retry-new-endpoints/proxy/round_tripper/proxy_round_tripper_test.go#L415-L523) which I think covers what you are pointing to.

I can't use your suggested solution [here](https://github.com/cloudfoundry/gorouter/pull/470#discussion_r2020684257) because it won't cover the case for when new instance is added and old is removed (the numEndpoints will be the same) and this is the case we are trying to solve when instance is being updated.

I added the [following logic](https://github.com/cloudfoundry/gorouter/blob/retry-new-endpoints/proxy/round_tripper/proxy_round_tripper.go#L155-L166): if the number of attempts will exceed the total number of endpoints we will check that the new selected endpoint is the one we have not tried yet, in that case we will try that new endpoint.

proxy/round_tripper/proxy_round_tripper.go
maxmoehl: How? 😄 Once the timestamps differ it can no longer stop and if all instances fail it will loop until MaxAttempts is hit.

proxy/round_tripper/proxy_round_tripper.go
mariash: @maxmoehl great! would you want to approve this PR?

tcdowney (approved):

maxmoehl (approved):
Code looks good and tests pass (after rebasing on main). Didn't get around to deploying it to one of our envs yet (but I'm fine with merging this).

PR #471, opened 2025-04-01
include gorouter_time metrics report

kart2bc:
- [ ] Read the [Contributing document](../blob/-/.github/CONTRIBUTING.md).

Summary
---------------

Backward Compatibility
---------------
Breaking Change? **No**

### DIFF
--- a/accesslog/schema/access_log_record.go
+++ b/accesslog/schema/access_log_record.go
@@ -124,6 +124,7 @@ type AccessLogRecord struct {
 	TlsHandshakeFinishedAt      time.Time
 	AppRequestFinishedAt        time.Time
 	FinishedAt                  time.Time
+	GorouterTime                float64
 
 	LocalAddress string
 }
@@ -134,17 +135,6 @@ func (r *AccessLogRecord) formatStartedAt() string {
 
 func (r *AccessLogRecord) roundtripTime() float64 {
 	return r.FinishedAt.Sub(r.ReceivedAt).Seconds()
-}
-
-func (r *AccessLogRecord) gorouterTime() float64 {
-	rt := r.roundtripTime()
-	at := r.appTime()
-
-	if rt >= 0 && at >= 0 {
-		return rt - at
-	} else {
-		return -1
-	}
 }
 
 func (r *AccessLogRecord) dialTime() float64 {
@@ -166,10 +156,6 @@ func (r *AccessLogRecord) tlsTime() float64 {
 		return -1
 	}
 	return r.TlsHandshakeFinishedAt.Sub(r.TlsHandshakeStartedAt).Seconds()
-}
-
-func (r *AccessLogRecord) appTime() float64 {
-	return r.AppRequestFinishedAt.Sub(r.AppRequestStartedAt).Seconds()
 }
 
 // failedAttemptsTime will be negative if there was no failed attempt.
@@ -281,7 +267,7 @@ func (r *AccessLogRecord) makeRecord(performTruncate bool) []byte {
 
 	// #nosec  G104 - ignore errors from writing the access log as it will only cause more errors to log this error
 	b.WriteString(`gorouter_time:`)
-	b.WriteDashOrFloatValue(r.gorouterTime())
+	b.WriteDashOrFloatValue(r.GorouterTime)
 
 	// #nosec  G104 - ignore errors from writing the access log as it will only cause more errors to log this error
 	b.WriteString(`app_id:`)
--- a/handlers/reporter.go
+++ b/handlers/reporter.go
@@ -57,6 +57,22 @@ func (rh *reporterHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request, ne
 		requestInfo.RouteEndpoint, proxyWriter.Status(),
 		requestInfo.ReceivedAt, requestInfo.AppRequestFinishedAt.Sub(requestInfo.ReceivedAt),
 	)
+	rh.calculateGorouterTime(requestInfo)
+	rh.reporter.CaptureGorouterTime(requestInfo.GorouterTime)
+}
+
+// calculateGorouterTime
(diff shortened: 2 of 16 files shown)


metrics/compositereporter.go
```diff
@@ -30,6 +30,7 @@ type MetricReporter interface {
 	CaptureRoutingRequest(b *route.Endpoint)
 	CaptureRoutingResponse(statusCode int)
 	CaptureRoutingResponseLatency(b *route.Endpoint, statusCode int, t time.Time, d time.Duration)
+	CaptureGoRouterTime(rt, at float64)
```
ameowlia, line 33:
```suggestion
	CaptureGorouterTime(rt, at float64)
```

:)

metrics/metricsreporter.go
```diff
@@ -77,6 +77,22 @@ func (m *Metrics) CaptureRoutingResponse(statusCode int) {
 	m.Batcher.BatchIncrementCounter("responses")
 }
 
+func (m *Metrics) CaptureGoRouterTime(rt, at float64) {
```
ameowlia, line 80: Let's not duplicate this logic. We should only calculate it once here: https://github.com/cloudfoundry/gorouter/blob/ffd637c4d77cf2639d68a9e1490134768f1a4f4b/accesslog/schema/access_log_record.go#L135-L148

handlers/reporter.go
```diff
@@ -57,6 +57,9 @@ func (rh *reporterHandler) ServeHTTP(rw http.ResponseWriter, r *http.Request, ne
 		requestInfo.RouteEndpoint, proxyWriter.Status(),
 		requestInfo.ReceivedAt, requestInfo.AppRequestFinishedAt.Sub(requestInfo.ReceivedAt),
 	)
+	appTime := requestInfo.AppRequestFinishedAt.Sub(requestInfo.AppRequestStartedAt).Seconds()
```
ameowlia, line 60: Gorouter has many [handlers](https://github.com/cloudfoundry/gorouter/blob/ffd637c4d77cf2639d68a9e1490134768f1a4f4b/proxy/proxy.go#L158-L194) that are called like an [onion](https://cloudfoundry.github.io/cf-networking-onboarding/gorouter/gorouters-have-layers-onions-have-layers).

The FinishedAt value, is set in the [access log handler](https://github.com/cloudfoundry/gorouter/blob/ffd637c4d77cf2639d68a9e1490134768f1a4f4b/handlers/access_log.go#L63). At this point in the code, the access log handler has not set that value.

ameowlia (changes requested):
I think there are at least two options forward:

1. Move the call to CaptureGorouter to the accesslog handler. And move the calculation to gorouter time to access log handler.
2. OR set "FinishedAt" in the reporter handler, calculate gorouter_time in the reporter handler, send a metric for gorouter_time in the reporter handler, set gorouter_time on the reqInfo in the reporter handler, and then use that calculated value for gorouter time in the access log handler.

handlers/reporter.go
kart2bc: Based on the sequence of calls, the accesslog handler been called before ReportHandler.
Hence, I thought the FinishedAt value should be available in the reportHandler.

metrics/metricsreporter.go
kart2bc: Absolutely! 
I would like to move this to the util package and make reporthandler and accesslogger to use this method.

handlers/reporter_test.go
```diff
@@ -103,6 +106,10 @@ var _ = Describe("Reporter Handler", func() {
 		Expect(latency).To(BeNumerically(">", 0))
 		Expect(latency).To(BeNumerically("<", 10*time.Millisecond))
 
+		Expect(fakeReporter.CaptureGorouterTimeCallCount()).To(Equal(1))
+		gorouterTime := fakeReporter.CaptureGorouterTimeArgsForCall(0)
+		Expect(gorouterTime).To(BeNumerically(">", 0))
```
ameowlia, line 111: Can you add a second test kind of like on line 107? To make sure it is a small value?

ameowlia:
I think you might also need to add a promethius metric implementation: https://github.com/cloudfoundry/gorouter/blob/main/metrics_prometheus/metrics.go

kart2bc:
* Test results *
``` cf query 'gorouter_time{source_id="gorouter"}'
{"status":"success","data":{"resultType":"vector","result":[{"metric":{"__v1_type":"ValueMetric","deployment":"cf-217600a306e78084eaf0","index":"e33e6db0-2af5-4b4a-adf7-8b18bb1ff447","ip":"10.10.0.12","job":"router","origin":"gorouter","product":"Small Footprint Tanzu Platform for Cloud Foundry","source_id":"gorouter","system_domain":"sys.tas.z547a4313.shepherd.lease"},"value":[1744150830.000,"0.26723100000000666"]}]}}```

File: accesslog/schema/access_log_record.go [hunk 5]
```diff
@@ -136,17 +137,6 @@ func (r *AccessLogRecord) roundtripTime() float64 {
...
-	rt := r.roundtripTime()
-	at := r.appTime()
-
-	if rt >= 0 && at >= 0 {
-		return rt - at
-	} else {
-		return -1
-	}
-}
-
```
geofffranks, line 149: I'm confused how the access log record tests still pass with no updates. After removing gorouterTime(), and having it log the pre-calculated GorouterTime value, the GorouterTime would be 0, unless the test was modified to set GorouterTime on the fake access log record used in the tests. 

... right? 😕

geofffranks (changes requested):

accesslog/schema/access_log_record.go
kart2bc: Figured this issue after running the whole suite. Addressed them here https://github.com/cloudfoundry/gorouter/pull/473.

PR #472, opened 2025-04-10
Fix route comparison check

maxmoehl:

Summary
---------------

A bunch of minor fixes to comparing endpoints when adding them to the route registry.

Backward Compatibility
---------------
Breaking Change? **No**


maxmoehl:
Turns out that gorouter was setting the new algo once the CC actually pushed it through but due to some subtle bugs it wasn't showing when dumping the routes. This PR now contains a bunch of small improvements to fix that and some other things as well as add some tests.

ameowlia (approved):

